PR #3, opened 2020-05-25
Port workflow conditional logic from idseq-web

kislyuk:
- Parameterize workflow inputs that were passed as "attributes" in idseq-dag and rendered as literals by idd2wdl
- Manually re-introduce conditional logic to handle single vs paired-end fastqs
- Manually re-introduce conditional logic to handle human host filtering when indicated host is not human
- Add a publish script that writes all workflow WDLs for every tag in the repo to s3://idseq-workflows

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,2 +1,5 @@
 lint:
 	find . -name '*.wdl' | xargs miniwdl check
+
+publish:
+	scripts/publish.sh
--- a/main/experimental.wdl
+++ b/main/experimental.wdl
@@ -103,6 +103,8 @@ task GenerateAlignmentViz {
     File taxid_annot_sorted_family_nr_fasta
     File taxid_locations_family_nr_json
     File taxid_locations_combined_json
+    String nt_db
+    String nt_loc_db
   }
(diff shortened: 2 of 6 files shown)


kislyuk:
* I had to switch to `Array[File]` because the files in the array have to be rendered as JSON input to the program, and the only way I could render them was using the following hack: `"{sep='","' input_array}"`. I would love it if you could show me a better way to do this (same goes for doing it without `sep`).

* Yes, the database files passed by their S3 URL are being downloaded by the code inside idseq-dag, and we should pull them out. You can see the full list in https://github.com/chanzuckerberg/idseq/pull/160/files#diff-c396655b3d3683aa7900dac3082e26caR90-R98.

* Yes, I absolutely want to provide those variables out-of-band - we discussed this previously - you said there is no good way to pass environment variables through to the task. One good way to do this, I think, would be to pass miniwdl an option like `--forward-env-vars AWS_DEFAULT_REGION DEPLOYMENT_ENVIRONMENT` which would get those variables from the environment and set them in all task containers.

PR #4, opened 2020-05-30
basic host_filter test

mlin:
@kislyuk here's a proposed test framework using [bash-tap](https://github.com/illusori/bash-tap) and WDL. I have not hooked up moto yet (at least the validation task reports it's lacking credentials, but doesn't actually fail itself). Will look at that next but let me know what you think of this in the meantime


mlin:
Oh, it depends on the monorepo Dockerfile. Would we relocate that into this repo eventually?

PR #5, opened 2020-06-06
start phasing out aws_region and deployment_env inputs

mlin:
@kislyuk proposing this way to phase out the aws_region and deployment_env inputs, which 'should' already be unnecessary, but something could still go wrong of course. Making them optional, so we can take them out of the inputs and later when we're satisfied they're no longer needed, delete them entirely.

The interpolation uses a [WDL trick](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#interpolating-and-concatenating-optional-strings) which evaluates `~{"foo" + x}` to the empty string if x is null.

### DIFF
--- a/main/experimental.wdl
+++ b/main/experimental.wdl
@@ -3,8 +3,8 @@ version 1.0
 task GenerateTaxidFasta {
   input {
     String docker_image_id
-    String aws_region
-    String deployment_env
+    String? aws_region
+    String? deployment_env
     String dag_branch
     String s3_wd_uri
     File taxid_fasta_in_annotated_merged_fa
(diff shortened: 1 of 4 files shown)


kislyuk:
Oh... I just pushed a branch that just deleted all occurrences of region and deployment_env. I don't see the need to tiptoe around it, IMHO we should get rid of them all and catch any issues in test.

kislyuk:
Hoping this isn't too flippant but I mean, this PR is adding lines where we are aiming to remove them :)

mlin:
@kislyuk Sure, I defer to your estimation of whether we'll need some transition period as we build out the unit test suite (which we should endeavor to make run without the non-portable plugin behaviors). I'm not really familiar with what/how things break without these env vars.

kislyuk:
- AWS_REGION can be set from the metadata supplied by ECS

- DEPLOYMENT_ENVIRONMENT is only used by remote non-host alignment, and will not be needed if that step is run locally

kislyuk:
Independently addressed in #6.

PR #9, opened 2020-06-18
build docker image through GitHub Actions+Packages

mlin:


mlin:
@kislyuk Thanks, the naming scheme & Dockerfile placement make more sense to me now. I have made a build matrix for now so that independent images can build in parallel. Will need to revisit if one builds on another.

PR #11, opened 2020-06-24
Add rapsearch2 to docker image

morsecodist:
Title says it all


main/Dockerfile
```diff
@@ -102,3 +102,21 @@ RUN curl -L -o /usr/local/bin/picard.jar https://github.com/broadinstitute/picar
 # Create a single executable so we can use SingleCommand
 RUN printf '#!/bin/bash\njava -jar /usr/local/bin/picard.jar "$@"\n' > /usr/local/bin/picard
 RUN chmod +x /usr/local/bin/picard
+
+# Install rapsearch2, required for running alignment locally
+RUN apt-get update && \
```
kislyuk, line 107: apt-get update already runs above, let's drop it from here. If the cache is bad, the build should automatically retry without cache.

main/Dockerfile
```diff
@@ -102,3 +102,21 @@ RUN curl -L -o /usr/local/bin/picard.jar https://github.com/broadinstitute/picar
...
+        apt-get install -y \
+                curl \
+                g++ \
+                # Runtime dependencies
+                # On versions higher than 1.62 compilation errors because of an overloading ambiguity with the `advance` method
+                libboost1.62-all-dev
```
kislyuk, line 113: Can we install only the libraries it needs?

morsecodist: You mean sub-libraries of boost?

morsecodist: I just tried to install a specific library `libboost1.62-chrono-dev` and it didn't exist. `liblibboost-chrono-dev` does install but unfortunately compiling breaks due to the issue I explain in the comment. It seems like we can't do specific installations with that specific version. To determine the version I tested every available version and selected the latest compatible one.

kislyuk: Sounds good, let's go with the simple solution of including all of it. It's a bit inefficient to include all of it, but there's no easy way to slim it down.

PR #17, opened 2020-07-07
enable more task tests by letting miniwdl handle small/medium S3 downloads

mlin:
The skeleton test cases for these three steps had been marked xfail while we work through public access to their reference databases on S3. Now miniwdl v0.8.0 can download public s3:// URIs without any external plugin.

* For small/medium downloads, changes the URI `String` inputs to `File` so that miniwdl takes over downloading
* the star and gsnap cases work, but the databases are too large to use in Actions or arguably for "unit tests" in general (27GB and 16GB). have marked them 'skip'
* because the auto-untar for star and gsnap host filter databases are no longer streaming, this diff would cause (moderate) speed and disk space usage regressions in production
* also skipped are tasks using nr/nt databases which are larger; leaving these handled by idseq-dag for the moment.
* eventually we may either store the databases under a S3 key prefix without tarring, or allow miniwdl's future implementation of the WDL 2.0 Directory type to accept tar inputs (which would be a useful but nonstandard feature)

### DIFF
--- a/.github/workflows/wdl-ci.yml
+++ b/.github/workflows/wdl-ci.yml
@@ -73,6 +73,6 @@ jobs:
 
           export DOCKER_IMAGE_ID="${IMAGE_URI}:${TAG}"
           # wip:
-          export IDSEQ_DAG_BRANCH=mlin-auto-local-untar
+          export IDSEQ_DAG_BRANCH=master
 
           make test-${{ matrix.workflow_dir }}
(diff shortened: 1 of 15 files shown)


.github/workflows/wdl-ci.yml
```diff
@@ -59,11 +59,20 @@ jobs:
...
           source /etc/profile
           sudo apt-get -qq update
           sudo apt-get -qq install --yes jq moreutils make virtualenv zip unzip httpie git shellcheck
           virtualenv --python=python3.6 .venv
           source .venv/bin/activate
           pip install -r requirements-dev.txt
+
           export DOCKER_IMAGE_ID="${IMAGE_URI}:${TAG}"
+          # wip:
+          export IDSEQ_DAG_BRANCH=master
```
kislyuk, line 76: Please remove L75-L76 before merging

PR #23, opened 2020-07-24
add miniwdl download cache cleaner script to main docker

mlin:
forward-ported from old monorepo PR.

### DIFF
--- a/main/Dockerfile
+++ b/main/Dockerfile
@@ -1,5 +1,6 @@
 FROM ubuntu:18.04
 ARG DEBIAN_FRONTEND=noninteractive
+ARG MINIWDL_VERSION=0.7.5
 
 LABEL maintainer="IDseq Team idseq-tech@chanzuckerberg.com"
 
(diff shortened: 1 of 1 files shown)


main/Dockerfile
```diff
@@ -51,11 +52,14 @@ RUN apt-get -q install -y \
 
 # Match versions from Ubuntu 20.04 LTS (please do not upgrade except to fix bugs)
 RUN pip3 install awscli==1.17.14 boto3==1.9.253
-RUN pip3 install miniwdl==0.7.5 miniwdl-s3parcp==0.0.5 miniwdl-s3upload==0.0.4
+RUN pip3 install miniwdl==${MINIWDL_VERSION} miniwdl-s3parcp==0.0.5 miniwdl-s3upload==0.0.4
 RUN pip3 install https://github.com/chanzuckerberg/miniwdl-plugins/archive/f0465b0.zip#subdirectory=sfn-wdl
 RUN pip3 install https://github.com/chanzuckerberg/idseq-dag/archive/v4.11.6.tar.gz
 RUN pip3 install https://github.com/chanzuckerberg/s3mi/archive/v0.8.0.tar.gz
 
+ADD https://raw.githubusercontent.com/chanzuckerberg/miniwdl/v${MINIWDL_VERSION}/examples/clean_download_cache.sh /usr/local/bin
```
mlin, line 60: Note: ADD in Dockerfile does not necessarily invalidate the build cache. It does always re-download the file which is a problem for big files (not here), but then it then checksums it to see if the step was cached.

mlin: https://github.com/moby/moby/issues/3672

PR #25, opened 2020-07-28
Add optional usegsnapl option

morsecodist:
Add optional option to set `gsnap` vs `gsnapl`. This allows us to default to `gsnap` locally and `gsnapl` remotely but still provides the option for a manual override.

### DIFF
--- a/main/non_host_alignment.wdl
+++ b/main/non_host_alignment.wdl
@@ -1,6 +1,6 @@
 version 1.0
 
-task RunAlignmentRemotely_gsnap_out {
+task RunAlignment_gsnap_out {
   input {
     String docker_image_id
     String dag_branch
@@ -44,7 +44,7 @@ task RunAlignmentRemotely_gsnap_out {
   }
 }
 
-task RunAlignmentRemotely_rapsearch2_out {
+task RunAlignment_rapsearch2_out {
   input {
     String docker_image_id
     String dag_branch
@@ -57,6 +57,7 @@ task RunAlignmentRemotely_rapsearch2_out {
     File? index
     String index_dir_suffix
     Boolean use_taxon_whitelist
+	Boolean? run_locally = false
   }
   command<<<
   set -euxo pipefail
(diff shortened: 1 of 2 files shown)


morsecodist:
Not sure why the test is failing. I don't really understand the error. Does anyone know?

main/non_host_alignment.wdl
```diff
@@ -14,6 +14,7 @@ task RunAlignmentRemotely_gsnap_out {
     String index_dir_suffix
     Boolean use_deuterostome_filter
     Boolean use_taxon_whitelist
+    Boolean? use_gsnapl
```
kislyuk, line 17: I think we need to set a default value for this - `true`?

mlin:
https://github.com/chanzuckerberg/idseq-workflows/pull/26 should fix the test. sorry

main/non_host_alignment.wdl
morsecodist: I currently have different behaviour for undefined vs false which is a little dicey I'll admit. What if I changed it to `force_gsnapl` and used no default so the false/undefined case is the default and leaves it up to the step itself. Then the true behaviour forces gsnapl then we can get rid of the no gsnapl remote behavior that throws an error.

kislyuk: OK - this is fine as long as it has actually been tested with `miniwdl run ... --input ...` where input contains all 3 scenarios.

kislyuk: Also would be great to document the possible values for this input (either in a comment preceding this line, or maybe using https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md#parameter-metadata)

mlin:
@morsecodist when the non_host_alignment.wdl gsnap task is stable to your satisfaction, we'll also want to port the new option to the [gsnap task in host_filter.wdl](https://github.com/chanzuckerberg/idseq-workflows/blob/75f05c2222f79747fd5bfa0f3eead4810743d4a5/main/host_filter.wdl#L360). I'll be glad to help with that

PR #27, opened 2020-07-31
Simplify CG docker image and update a few packages

tfrcarvalho:
* Upgrade to match biohub's image
* Use base conda environemnt


consensus-genome/Dockerfile
```diff
@@ -16,19 +16,15 @@ RUN curl -sSL https://repo.anaconda.com/miniconda/Miniconda3-py37_4.8.3-Linux-x8
   && bash /tmp/miniconda.sh -bfp /usr/local \
   && rm -rf /tmp/miniconda.sh \
   && conda update conda \
-  && conda install -y python=3.6 \
+  && conda install -y python=3.7.4 \
```
jshoe, line 19: do you know why they picked 3.7.4? instead of 3.8 maybe

tfrcarvalho: There are a lock of packages with strict dependencies. I couldn't make it work with 3.8.

PR #28, opened 2020-08-07
Add "no reads left" errors

tfrcarvalho:
### Description

* Add error handling for no reads left after `RemoveHost`, `FilterReads` and `TrimReads` steps.
* Remove conditional execution. It will now return an error instead of skip steps.

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -1,4 +1,5 @@
-# The following pipeline was imported from previous work at: https://github.com/czbiohub/sc2-illumina-pipeline
+# The following pipeline was initially based on previous work at: https://github.com/czbiohub/sc2-illumina-pipeline
+# workflow version: consensus-genomes-1.1.0
 version 1.0
 
 workflow consensus_genome {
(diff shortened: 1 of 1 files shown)


consensus-genome/run.wdl
```diff
@@ -1,4 +1,5 @@
-# The following pipeline was imported from previous work at: https://github.com/czbiohub/sc2-illumina-pipeline
+# The following pipeline was initially based on previous work at: https://github.com/czbiohub/sc2-illumina-pipeline
+# workflow version: consensus-genomes-1.1.0
```
tfrcarvalho, line 2: @kislyuk, I was wondering if we can use some kind of convention to set the version in the workflow file from which we could create tags automatically.
This particular method does not work for multi-stage. So maybe we should have a file on the workflow folder with version + release notes?

consensus-genome/run.wdl
```diff
@@ -267,9 +270,9 @@ task QuantifyERCCs {
 }
 
 task FilterReads {
-    # TODO: process errors: no reads left
     input {
         String prefix
+        # SARS-CoV-2 default
```
katrinakalantar, line 275: great comment!

PR #29, opened 2020-08-11
remove minDepth param for pileup and nonessential steps

katrinakalantar:
To ensure consistency in major updates to the CG workflow, this PR includes changes parallel to this [PR](https://github.com/czbiohub/sc2-illumina-pipeline/pull/58). In particular, "The minimum depth parameter in samtools pileup is approximate since it discards whole reads (rather than applying the truncation separately at each position the read spans). This appears to especially cause bad behavior for some amplicon library preps, where overlapping regions between amplicons can cause nearly all reads to be discarded at the 5' end of an amplicon."

Additionally, we remove two pipeline steps that are extraneous to the primary pipeline goal of generating consensus genomes - `RealignConsensus` and `IntrahostVariants` - to streamline the pipeline.

The changes have been tested on five samples (ARTIC v3 and MSSPE). Results are as expected.

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -26,14 +26,9 @@ workflow consensus_genome {
         # could they be extracted to a separate params file? put in .json input to wdl?
         Boolean trim_adapters = true
 
-        Float   intrahost_min_frac = 0.5
-        String  intrahost_ploidy   = "2"
-        Boolean intrahost_variants = true
-
         Float ivarFreqThreshold = 0.9
         Int   ivarQualTreshold  = 20
         Int   minDepth          = 10
-        Int   mpileupDepth      = 10000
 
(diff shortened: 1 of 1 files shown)


consensus-genome/run.wdl
```diff
@@ -27,14 +27,9 @@ workflow consensus_genome {
         # could they be extracted to a separate params file? put in .json input to wdl?
```
jshoe, line 27: Should we bump the version up top, Tiago?

tfrcarvalho: An official version was not released yet. Will probably release both changes at the same time.
But we can make it `1.2.0` anyway, in case we need to revert this one by any chance.
(for now the comment version is just a reference and the version is actually defined by a tag).

Can you make that change @katrinakalantar ?

PR #33, opened 2020-08-14
Rename main workflow to short-read-mngs

kislyuk:


kislyuk:
I think duplicating will actually lead to more issues, since the code will then be duplicated and may diverge, with further unintended effects.

The test strategy for this is to merge this here and then test the release process in the dev environment before moving on to staging.

tfrcarvalho:
> since the code will then be duplicated and may diverge,

The idea of duplicating was that the old name would be frozen in time. Not to keep them in sync. 
My issue was not so much the testing of `idseq` managed elements, but letting `idseq-web` switch from one infra to the other independently.

But I guess the old workflow/s3 keys are not removed when you publish, which will allow the old workflow to still be used, right?

kislyuk:
Correct - the published artifacts (WDLs on S3 and Docker images) are persistent. This allows our code to evolve while old builds remain reproducible. It is essential for us to have a strategy that does not force us to freeze parts of the codebase to accommodate legacy code.

PR #34, opened 2020-08-14
Import idseq-dag contents

kislyuk:


short-read-mngs/idseq-dag/examples/amr_human_paired_sample_with_rpm.json
```diff
@@ -0,0 +1,31 @@
+{
```
tfrcarvalho, line 1: Does it make sense to migrate these json examples?

kislyuk:
The strategy here is to retire the idseq-dag repository (using the github "archive" functionality to make it read-only) and continue any necessary idseq-dag development here. We can certainly remove obsolete code later, but for now I removed the bare minimum of files that were present in the old repo, to preserve the full history and maintain as much continuity between the repos as possible. Even though the `idseq_dag/engine/pipeline_flow.py` file and other files are no longer in use, I think they are still imported by some of the code paths.

short-read-mngs/idseq-dag/examples/amr_human_paired_sample_with_rpm.json
kislyuk: I would like to keep these examples for now to maintain continuity (this import is not intended to cull all obsolete files from the idseq-dag codebase). We can later remove these files in a separate PR while also considering if we need to add any replacement functionality in the new workflow infrastructure.

tfrcarvalho:
OK, I have a personal preference to migrate selectively even if it requires more testing and might code to be adapted.
But I can see the pros of your approach as a bit safer.

kislyuk:
Manually merged in 628fb331fd3612663dbf89e5018d05c31d80ae06 (github is unable to merge this PR correctly in the UI).

PR #37, opened 2020-08-17
add gsnap/gsnapl exe detection to host filter

mlin:
Copy @morsecodist technique developed for non-host alignment stage


mlin:
Subsumed by #39

PR #39, opened 2020-08-19
local driver wdl for short-read-mngs

mlin:
Adding a top-level WDL to schedule the four stages locally, and various minor adjustments getting that to work.


short-read-mngs/idseq-dag/idseq_dag/steps/run_alignment.py
```diff
@@ -199,13 +199,13 @@ def run_locally(self, input_fas, output_m8):
...
             error_message = run(
-                ['gsnapl-2018-10-26', '-D', index_path, '-d', self.genome_name],
+                ['gsnapl.avx2-2018-10-26', '-D', index_path, '-d', self.genome_name],
                 input='>'.encode('utf-8'),
                 stderr=PIPE,
                 stdout=PIPE
             ).stderr
             # note, alignment uses a pinned version of gmap/gsnap
-            gsnap_command = "gsnap-2018-10-26" if 'please run gsnap instead' in error_message.decode('utf-8') else "gsnapl-2018-10-26"
+            gsnap_command = "gsnap.avx2-2018-10-26" if 'please run gsnap instead' in error_message.decode('utf-8') else "gsnapl.avx2-2018-10-26"
```
mlin, line 208: @morsecodist FYI, one more subtlety I found in the index format saga. The `gsnap[l]` executable is just a dispatcher that runs `gsnap[l].avx2` or one of the others, depending on CPU architecture. But `gsnap[l]-2018-10-26` wouldn't know to run `gsnap[l].avx2-2018-10-26`, it would invoke the older version `gsnap[l].avx2`. So we can invoke `gsnap[l].avx2-2018-10-26` directly here. I was able to use the k=16 index generated with that version this way. 😭

morsecodist: I originally did the version mapping thinking this was the cause of the local failure. It turned out to be the k-mer thing. It is more rigorous to use the same version but we could probably abandon this approach and use the version from apt if we wanted.

mlin: @morsecodist To clarify, the k=16 indices generated with gmap_build 2018-10-26 do work with this diff. The problem was that the gsnap-2018-10-26 entry point would actually dispatch to the old, 2017 gsnap.avx2 executable, which wasn't able to read the k=16 index generated by the newer gmap_build. The above change ensure we use the desired gsnap.avx2-2018-10-26 executable with the matched index version.

short-read-mngs/experimental.wdl
```diff
@@ -308,13 +312,17 @@ workflow idseq_experimental {
...
 
-  call RunSRST2 {
-    input:
-      docker_image_id = docker_image_id,
-      dag_branch = dag_branch,
-      s3_wd_uri = s3_wd_uri,
-      fastqs = select_all([fastqs_0, fastqs_1]),
-      file_ext = file_ext
+  if (defined(resist_genome_db) && defined(resist_genome_bed)) {
+    call RunSRST2 {
```
mlin, line 316: Make running this step optional since it uses non-public reference databases. We will need to update the inputs provided by backend to supply these S3 paths.

short-read-mngs/idseq-dag/idseq_dag/steps/run_gsnap_filter.py
```diff
@@ -55,6 +56,14 @@ def run(self):
                                      auto_untar=True)
         gsnap_base_dir = os.path.dirname(genome_dir)
         gsnap_index_name = os.path.basename(genome_dir)
+        # Hack to determine gsnap vs gsnapl
```
mlin, line 59: Port Todd's technique developed for non-host alignment into this host filter stage.

mlin:
@kislyuk @morsecodist This PR contains the changes/additions needed to run short-read-mngs locally, including a top-level "driver" WDL and boilerplate input YAML files. There are a few important things to go over together:

1. Transitions `nr` and `nt` downloads in postprocess & experimental to be managed by miniwdl instead of idseq-dag (so that locally, miniwdl can download it without AWS credentials). We'll want to monitor SFN-WDL carefully with this, because it makes the miniwdl download cache (and cleanup process for it) much more critical than it is now.
2. Moves away from the pattern where default reference database URIs were being set by interpolation from other input variables. The `File an_input = "<uri>"` construct is a rather grey area of the WDL spec; I'm not sure if it works on other WDL runners even if URI is a literal, let alone with interpolations too. It's kind of accidental that it works in miniwdl, and was causing some issues with the download caching. I'll open a WDL ticket about clarifying this, in the meantime striking the interpolations   at least avoids current miniwdl problems.
3. Minor modifications needed to the inputs IDseq provides for RunSRST2 -- I'll look into this in parallel

tests/short-read-mngs/local_test_viral.yml
```diff
@@ -0,0 +1,29 @@
...
+non_host_rapsearch2_index: s3://idseq-database/test/viral-alignment-indexes/viral_nr_rapsearch.tar.gz
+non_host_gsnap_index: s3://idseq-database/test/viral-alignment-indexes/viral_nt_gsnap2018-10-26_k16.tar.gz
+non_host_gsnap_genome_name: viral_nt_gsnap2018-10-26_k16
+postprocess.nt_db: s3://idseq-database/test/viral-alignment-indexes/viral_nt
+postprocess.nt_loc_db: s3://idseq-database/test/viral-alignment-indexes/viral_nt_loc.db
+postprocess.nr_db: s3://idseq-database/test/viral-alignment-indexes/viral_nr
+postprocess.nr_loc_db: s3://idseq-database/test/viral-alignment-indexes/viral_nr_loc.db
+experimental.nt_db: s3://idseq-database/test/viral-alignment-indexes/viral_nt
+experimental.nt_loc_db: s3://idseq-database/test/viral-alignment-indexes/viral_nt_loc.db
+experimental.nt_info_db: s3://idseq-database/test/viral-alignment-indexes/viral_nt_info.db
```
kislyuk, line 29: Where are these .yml files consumed/used?

short-read-mngs/idseq-dag/idseq_dag/steps/run_alignment.py
morsecodist: Wait so a k16 mini index does work here? That's awesome.

tests/short-read-mngs/local_test_viral.yml
mlin: In the tutorial the user is directed to pass them into `miniwdl run`. Planning to add the viral one to CI once we deal with the 64GB accessions2taxid.db.

PR #41, opened 2020-08-25
switch to a warning for multiple emitted reads from the same cluster

morsecodist:
Switch this assertion with a warning so it doesn't crash the whole pipeline.

### DIFF
--- a/short-read-mngs/idseq-dag/idseq_dag/util/cdhit_clusters.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/util/cdhit_clusters.py
@@ -30,6 +30,7 @@ deduplicated read is not the one marked with '*'.  Hence we handle that case cor
 and put a lot of assertions to make sure problems with cdhit output are detected.
 """
 
+import idseq_dag.util.log as log
 from idseq_dag.util.fasta import iterator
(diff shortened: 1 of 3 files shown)


short-read-mngs/idseq-dag/idseq_dag/util/cdhit_clusters.py
```diff
@@ -67,17 +68,16 @@ def record_cluster_size(
...
+        if len(emitted_reads_from_cluster) != 1:
+            """
+            If this check is true, cd-hit-dup has emitted multiple reads
+            from the same cluster. We have observed that this does happen.
+            Everything will run fine, but read counts contributed by that
+            cluster will be exaggerated. We are in the process of replacing
+            cd-hit-dup with a tool that does not have this bug but until
+            that time we will just output a warning so we don't block results.
+            """
+            log.write(f"more than one read emitted from cd-hit-dup cluster: {emitted_reads_from_cluster} reads emitted on line {line_number}", warning=True)
```
cdebourcy, line 80: Would it be possible to alert on this? I doubt we'll actually look at these logs... but we should definitely know when this is happening and potentially even inform the user

PR #44, opened 2020-09-01
 add short-read-mngs end-to-end test to CI (using viral databases)

mlin:
This first cut tests that the workflow runs successfully but doesn't include output correctness checks.


mlin:
@morsecodist @katrinakalantar @cdebourcy With the viral databases, we can now run the mngs workflow from beginning to end in GitHub Action, which is great.

Since it takes only 10-15 minutes, I'm actually thinking of replacing the task-level unit tests from earlier in the summer, since the latter have a lot more maintenance overhead for test fixtures etc. A little further work is needed so that we can still write test assertions on intermediate results in the workflow.

PR #45, opened 2020-09-01
Emit depths file for consensus genome.

tfrcarvalho:
### Description

* Output depths file to be used in coverage visualization results.


consensus-genome/run.wdl
```diff
@@ -1,5 +1,5 @@
 # The following pipeline was initially based on previous work at: https://github.com/czbiohub/sc2-illumina-pipeline
-# workflow version: consensus-genomes-1.2.2
+# workflow version: consensus-genomes-1.2.4
```
jshoe, line 2: + 0.0.2 ?

tfrcarvalho: We forgot to update it... and there is already a 1.2.3.
This is just a comment for information and it is not used for the release. We could just remove it.

PR #48, opened 2020-09-26
Add support for single-end fastqs

tfrcarvalho:
### Description

* Support single-end fastqs (in addition to paired-end): [CH-18084](https://app.clubhouse.io/idseq/story/18084)
* Fixes bug where the FilterReads step would error earlier with no reads left without reporting the correct error
* Fixes bug where non gzip input fastqs would originated 0 total reads
* Improves precision of quantifiers for `Array[File]` outputs

### Tests

* Tested locally with:
  * Paired-end Gzip Fastq with no reads left after in FilterReads task
  * Single-end Gzip Fastq with consensus genome created
  * Paired-end Gzip Fastq with consensus genome created (same output as single-end)
  * Paired-end uncompressed Fastq with consensus genome created

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -6,7 +6,7 @@ workflow consensus_genome {
     input {
         # Required parameters
         File fastqs_0
-        File fastqs_1
+        File? fastqs_1
 
         String docker_image_id
         File ercc_fasta
@@ -44,8 +44,7 @@ workflow consensus_genome {
     call RemoveHost {
         input:
             prefix = prefix,
-            fastqs_0 = fastqs_0,
-            fastqs_1 = fastqs_1,
+            fastqs = select_all([fastqs_0, fastqs_1]),
             ref_host = ref_host,
             docker_image_id = docker_image_id
     }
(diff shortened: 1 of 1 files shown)


consensus-genome/run.wdl
```diff
@@ -139,41 +133,38 @@ workflow consensus_genome {
...
-                Quast.quast_tsv,
-                AlignReads.alignments,
-                QuantifyERCCs.ercc_out,
-                QuantifyERCCs.ercc_out,
-                ComputeStats.output_stats,
-                CallVariants.variants_ch
-            ]),
+            outputFiles = select_all(flatten([
+                RemoveHost.host_removed_fastqs,
+                select_all([
```
jshoe, line 146: just curious, do you need the additional `select_all` inside if you're flattening the whole?

jshoe (approved):
LGTM with a question

consensus-genome/run.wdl
tfrcarvalho: This was a whole fight with miniwdl type checking... so, yes, but i not 100% sure on an explanation.
Might follow up on that with @mlin later.

tfrcarvalho: I believe it was something to do with some of those files being optional...

PR #51, opened 2020-10-09
Compute and export alignment results from combining NT and NR hits

tfrcarvalho:
### Description

* Adds new step for computing a single classification based on both NT and NR results for both per read taxon counts and contig statistics
* Simplifies `CombineTaxonCounts` to enable using it for combining the extra taxon counts
* Changes rapsearch to output e-values in raw format (consistent with gsnap) - important for merging  and avoid weird conditionals  
* Adjust average alignment length to use one nucleotide as reference
* Bug fix: Taxon blacklist was not having any effect since we were reading from `additional_attributes` instead of `additional_files`.


short-read-mngs/Dockerfile
```diff
@@ -139,4 +136,7 @@ RUN make
 ENV PATH="${PATH}:/rapsearch2/Src/"
 # Uninstall build only dependencies
 RUN apt-get purge -y g++ libperl4-corelibs-perl make
-WORKDIR / 
+WORKDIR /
+
+COPY idseq-dag /tmp/idseq-dag
+RUN pip3 install /tmp/idseq-dag && rm -rf /tmp/idseq-dag
```
tfrcarvalho, line 142: This help to accelerate editing and building the image to run the step locally... any issues with this?

short-read-mngs/idseq-dag/idseq_dag/steps/compute_merged_taxon_counts.py
```diff
@@ -0,0 +1,127 @@
...
+                    self._write_tsv_row(nr_hit_dict, TAB_SCHEMA_MERGED, output_hit)
+
+        print("LAST STEP")
+        # Create new merged m8 and hit summary files
+        self.create_taxon_count_file(merged_m8_filename, merged_hit_filename, cluster_sizes_filename, merged_taxon_count_filename)
+
+    def create_taxon_count_file(self, merged_m8_filename, merged_hit_filename, cluster_sizes_filename, merged_taxon_count_filename):
+        # TOOO: Can this be consolidated throughout the pipeline?
+        # This setup is mostly repeated in three steps. The list of taxa do not seem to change.
+        evalue_type = 'raw'
```
tfrcarvalho, line 91: We cannot just use `raw` since rapsearch e-value is return log-based. Considering moving it inside the combine function...

short-read-mngs/idseq-dag/idseq_dag/steps/compute_merged_taxon_counts.py
```diff
@@ -0,0 +1,126 @@
...
+    def run(self):
+        # Retrieve inputs and outputs
+        nt_m8, nt_hitsummary2_tab, nr_m8, nr_hitsummary2_tab, cluster_sizes_filename = self.input_files_local
+        merged_m8_filename, merged_hit_filename, merged_taxon_count_filename = self.output_files_local()
+
+        # Create new merged m8 and hit summary files
+        nr_alignment_per_read = {}
+        # TODO: consider going through nt first, since that will save us from storing results in memory
+        # for all the reads that get they hit from NT contigs
+        for nr_hit_dict in parse_tsv(nr_hitsummary2_tab, TAB_SCHEMA_MERGED, strict=False):
```
tfrcarvalho, line 29: This a memory efficient solution... but if we guarantee enough memory we should be able to to accelerate it.
What has been our strategy until now?

tfrcarvalho: Files seem to be ~200MB in size (worst case). Will probably read all into memory.

short-read-mngs/Dockerfile
morsecodist: None, I thought to do something similar even

short-read-mngs/idseq-dag/idseq_dag/steps/run_alignment.py
```diff
@@ -638,6 +638,7 @@ def step_description(self, require_docstrings=False):
                 -l 10
                 -a T
                 -b 0
+                -s f
```
morsecodist, line 641: Can you explain this in a comment somewhere?

short-read-mngs/idseq-dag/idseq_dag/util/m8.py
```diff
@@ -71,17 +71,24 @@
...
             try:
                 row = raw_line.rstrip("\n").split("\t")
-                assert len(row) == len(schema)
+                if strict:
+                    assert len(row) == len(schema)
+                length = min(len(row), len(schema))
                 row_dict = {cname: ctype(vstr) for vstr,
-                            (cname, ctype) in zip(row, schema_items)}
+                            (cname, ctype) in zip(row[0:length], schema_items[0:length])}
             except:
```
morsecodist, line 92: Tiny nitpick but I think it would be a little cleaner to write this in the if so there is no indirection when reading the code. At first glance, I would assume this function would protect me from bad schemas but I see it just logs a warning. Maybe something like:

```python
if strict and len(row) != len(schema):
  msg = f"{path}:{line_number + 1}:  Parse error.  Input does not conform to schema: {schema}"
  log.write(msg, warning=True)

File: short-read-mngs/idseq-dag/idseq_dag/util/schemas.py [hunk 6]
```diff
@@ -0,0 +1,19 @@
...
+    "contig_accession_id": str,
+    "contig_species_taxid": int,
+    "contig_genus_taxid": int,
+    "contig_family_taxid": int,
+    "from_assembly" : bool,
+}
+
+TAB_SCHEMA_MERGED = dict(TAB_SCHEMA, **{
+    "source_count_type": str
+})
```
morsecodist, line 19: newline

short-read-mngs/non_host_alignment.wdl
```diff
@@ -83,22 +83,15 @@ task CombineTaxonCounts {
...
     String docker_image_id
     String s3_wd_uri
-    File gsnap_m8
-    File gsnap_deduped_m8
-    File gsnap_hitsummary_tab
-    File gsnap_counts_with_dcr_json
-    File rapsearch2_m8
-    File rapsearch2_deduped_m8
-    File rapsearch2_hitsummary_tab
-    File rapsearch2_counts_with_dcr_json
```
morsecodist, line 93: I worry about this a bit. Now we have an order dependency from the step that is not captured in the step itself.

short-read-mngs/idseq-dag/idseq_dag/steps/compute_merged_taxon_counts.py
morsecodist: In general, I tend more towards memory rather than speed. The pipeline runtime is dominated by a few costly tools so usually, I favor reliability over the speed of relatively fast operations. That said 200MB isn't so bad so up to you.

tfrcarvalho: ok, will leave as is... and optimize differently once/if it becomes a bottleneck

short-read-mngs/idseq-dag/idseq_dag/steps/run_alignment.py
tfrcarvalho: Dropping this. The format of e-value becomes very weird... a mix of float with hex like `0x1.4f41d47a6d132p-57`.

short-read-mngs/idseq-dag/idseq_dag/util/m8.py
tfrcarvalho: Sure, i was keeping the previous pattern but that sound better to me.

short-read-mngs/non_host_alignment.wdl
tfrcarvalho: As I had the chance to explain offline, the order is not important. Set was using index [3] to select the relevant file from each group. The inputs were kept as a basic refactor ignoring the contents of the step during the initial conversion to WDL.
So, now we send only the relevant file as input.

short-read-mngs/idseq-dag/idseq_dag/steps/compute_merged_taxon_counts.py
```diff
@@ -0,0 +1,124 @@
...
+                    output_m8.write(nt_m8_row)
+                    nt_hit_dict["source_count_type"] = "NT"
+                    self._write_tsv_row(nt_hit_dict, TAB_SCHEMA_MERGED, output_hit)
+                    if nr_alignment:
+                        del nr_alignment_per_read[nt_hit_dict["read_id"]]
+                elif nr_alignment and nr_alignment.contig:
+                    continue
+                elif nt_alignment.read:
+                    output_m8.write(nt_m8_row)
+                    nt_hit_dict["source_count_type"] = "NR"
```
katrinakalantar, line 59: should `source_count_type` here == `NT`?

short-read-mngs/idseq-dag/idseq_dag/util/m8.py
```diff
@@ -507,13 +516,16 @@ def generate_taxon_count_json_from_m8(
                 assert m8_line_columns[0] == hit_line_columns[0], msg
                 percent_identity = float(m8_line_columns[2])
                 alignment_length = float(m8_line_columns[3])
+                if count_type == 'NR' or hit_source_count_type == 'NR':
```
katrinakalantar, line 519: Can we add a comment to indicate why the `alignment_length *= 3` for future reference? 

i.e. Protein alignments (NR) are done at amino acid level. Each amino acid is composed of 3 nucleotides. To make alignment length values comparable across NT and NR alignments (for combined statistics), the NR alignment lengths are multiplied by 3. 

related q: will the affect the alignment lengths outside of the NT/NR combined prototype?

tfrcarvalho: Yes, is that ok?

tfrcarvalho: I can keep it just within the protoype for now if you prefer...

short-read-mngs/idseq-dag/idseq_dag/steps/compute_merged_taxon_counts.py
tfrcarvalho: Great catch!

short-read-mngs/idseq-dag/idseq_dag/util/m8.py
katrinakalantar: It would be ideal to keep this within the prototype for now if possible - it runs a risk of confusing people to change in the existing report. Additionally, since the current report separates the values it is possible to tease apart the differences and understand that NR `L`s would be lower.

tfrcarvalho: Done

katrinakalantar (commented):
This looks good to me - thank you for adding in the contig-merging capabilities.

morsecodist (approved):

PR #52, opened 2020-10-09
fix reading header bug

morsecodist:
Uncovered thanks to the unit tests in the idseq monorepo. This bug adds `representative read id` to the clusters dictionary because it doesn't skip the csv header. It doesn't actually break anything in practice because this ID is never looked up but it ends up in the cluster sizes file as a cluster.


tfrcarvalho (approved):
Did the tests fail on previous PR? Is it not a blocking check?

morsecodist:
@tfrcarvalho For some reason, the wdl tests are in the idseq monorepo PR and won't run on this codebase unless it's merged. I am not quite sure why I mentioned it to @mlin, maybe I did something wrong. It got through my manual testing because it doesn't affect the pipeline at all other than the random extra row in cluster sizes.

PR #54, opened 2020-10-18
local alignment: extract db to temp space instead of output dir

mlin:
Improves efficiency when miniwdl is using a shared filesystem for working directories, per https://miniwdl.readthedocs.io/en/latest/runner_advanced.html#write-scratch-files-under-tmpdir


mlin:
@morsecodist incremental efficiency improvement for the local indexes here. The Directory feature will be available in the upcoming miniwdl v0.9.x series

morsecodist (approved):
Good improvement, thanks

PR #55, opened 2020-10-19
Fix file reference to use miniwdl download caching logic

tfrcarvalho:
### Description

For the `short-read-mngs` pipeline, found just one misconfiguration leading to a download from `idseq-dag` code.

Note that are a few events of `fetch_from_s3` in the logs. Those seem to come all from the phylo tree pipeline.

### Test

Tested by doing a local run. All files were download by miniwdl.


jshoe (approved):

morsecodist (approved):
Thanks for finding this!

cdebourcy:
Wow!

PR #56, opened 2020-10-20
Bug Fix: fixes logical condition when applying length correction

tfrcarvalho:
We were multiplying both NT and NR values by 3.


katrinakalantar (approved):
This looks good to me. It will only adjust lengths for NR alignments - thank you for finding and making the change!

PR #57, opened 2020-10-23
Fix glob on trim reads that would not catch all output files

tfrcarvalho:
* Fix glob on trim reads that would not catch all output files


katrinakalantar (commented):
##This looks good to me!

katrinakalantar (approved):

PR #60, opened 2020-10-31
Stop filtering short reads in validate input and make count more reliable

tfrcarvalho:
Ticket: [CH-18285](https://app.clubhouse.io/idseq/story/18285)

* Stop filtering short reads in `validate_input` step. This reads will most likely be remove further downstream and they were not consistently removed from paired-end fastqs (see discussion in the ticket).
* Make count more reliable instead of relying on a parallel data structure


cdebourcy (approved):
Lgtm

PR #62, opened 2020-11-30
idseq_dag/util/fasta.py: don't mangle FASTQ read names with pipe characters

mlin:
Change intermediate delimiter used by `util.fastq.sort_fastx_by_entry_id()` from | to ` in hopes the latter is even less likely to arise in read names, causing the output to be mangled. (I'm working with some CAMI challenge datasets that use pipes in their read names)

### DIFF
--- a/short-read-mngs/idseq-dag/idseq_dag/util/fasta.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/util/fasta.py
@@ -78,10 +78,10 @@ def sort_fastx_by_entry_id(fastq_path):
         with open(tmp_sorted_path, 'wb') as out_file:
             # Command based on this https://www.biostars.org/p/15011/#103041
             if input_file_type(fastq_path) == 'fastq':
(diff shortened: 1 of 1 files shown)


tfrcarvalho (approved):
I guess it sounds rarer so seem fine, but still weird effects if it occurs.
Would you recommend a larger, more "bullet-proof" refactor of this operation?

mlin:
Thanks @tfrcarvalho -- ideally yea, but I think it's a small risk probably not worth spending the time to re-engineer anytime soon. Pipe isn't too unusual of a character to see used as a delimited "in the wild" (NCBI uses it a lot for example). Famous last words, but we'll probably be okay with backtick ;)

mlin:
LOL, and 5 minutes later I remembered that backtick can appear in the FASTQ [quality scores](https://en.wikipedia.org/wiki/FASTQ_format#Encoding) for older Illumina data. According to the diagram there, even pipe can appear there for PacBio data!

Maybe we should fix this properly after all, or see if it'll work with a nonprintable character. Reverted for now

PR #65, opened 2020-12-11
Low hanging fruit optimization of cluster file parsing

morsecodist:
This parsing step was made WAY slower (so slow it times out) by how I set up the tuples. I knew it would be bad but not that bad. idseq-dedup doesn't output the clusters in order so I couldn't create the tuples once like we did with cd-hit-dup. After some research I determined there is no particular reason for these to be tuples and appending to a list makes it even faster than it used to be. I can make this even faster by updating idseq-dedup to output files that are even more compatible with our pipeline though that probably isn't super necessary if we merge this in.

I also removed some dead caching code. We used to cache the result of `load_duplicate_cluster_sizes` so we could re-use it within a stage, we never call it more than once within a step. Now that our steps our isolated in miniwdl this cache will never be hit.

### DIFF
--- a/short-read-mngs/idseq-dag/idseq_dag/steps/nonhost_fastq.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/steps/nonhost_fastq.py
@@ -1,6 +1,6 @@
 import os
 
-from typing import Dict, Optional, Sequence, Set, List
+from typing import Dict, Optional, Sequence, Set, List, Tuple
 
 import idseq_dag.util.command as command
 import idseq_dag.util.command_patterns as command_patterns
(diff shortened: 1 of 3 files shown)


short-read-mngs/idseq-dag/idseq_dag/steps/nonhost_fastq.py
```diff
@@ -1,6 +1,6 @@
 import os
 
-from typing import Dict, Optional, Sequence, Set, Tuple
```
mlin, line 3: @morsecodist look like you still need Tuple here on L97

morsecodist: Thanks for catching this

PR #67, opened 2020-12-18
[CH-27791] Add check for "No reads after MakeConsensus"

jshoe:
- Tickets: https://app.clubhouse.io/idseq/story/27791/bug-on-computestats-step-of-consensus-genome-pipeline

# Description
- This should address the issues noted in the bug. We just need to check if the `consensus.fa` file has any actual reads.

# Tests
- Tested directly on an EC2-IA instance in production. I tested 3 of the samples noted in the ticket and successfully got the error message.

Screen Shot 2020-12-17 at 6 36 34 PM

- As a sanity check, I also ran a "normal sample" and it was successful.


tfrcarvalho (approved):
Looks good! Did you do a run? Or will check on staging?

jshoe:
I'm going to do a local run after merging/releasing, then staging

jshoe:
Worked as expected in local run

Screen Shot 2020-12-18 at 5 27 51 PM

PR #68, opened 2020-12-21
Parsing Refactor

morsecodist:
This one's a big one but I am very pleased with how it came out. Still adding some tests but I wanted to flag this as ready for review so we can get the ball rolling.

This fixes some problems in the codebase:
- The codebase had a lot of custom tsv parsing. This is a problem because that module handles edge cases nicely.
- The codebase had multiple reusable custom parsers for the same file types. In the case of `iterate_m8` it contained specific business logic in addition to parsing which wasn't immediately clear.
- There were numerous examples of ad-hoc parsing inside particular functions, despite having parsers for them implemented. This increased the complexity of those functions dramatically and it is unclear what file format they expect.
- In general, it was extremely difficult to know the format of any file in advance. There were many variants of each file and the schemas were not necessarily strictly enforced and it was unclear why. It was very challenging to determine which fields you could depend on and which you couldn't.
- The old schema system relied on dictionary key order, which technically works but isn't standard
- There were numerous examples of list order and length being used to hold different data. This made it difficult to track which data was in a particular data structure when it was passed across and around files.
- Refactoring parsers to a different file format would have been an extreme challenge due to how heterogenous everything was. (It was a big challenge even with the same format)

There were probably other issues as well but these are the major ones. I know this was a big lift but I really felt this needed to change.

To solve this problem I created a wrapper around python's csv parsers to add support for schemas. I would like to standardize the file formats, add headers, and explicitly add null values in any missing fields but I wanted to leave the file formats themselves unchanged here. So I needed a solution for multiple slightly different schemas. To address this I created a system of different schema variants. So instead of making every value optional, as we did before in some cases, rows must now conform to one of a set of variants. This means we can guarantee the presence of certain values in certain situations, and it is easier to reason about the types of files we may encounter. You can read more about it in the `parsing.py` file.

This does make the code a bit slower. However, I firmly believe this is worth it. The python code is usually not the bottleneck of our pipeline runs. I think reliable maintainable code is the priority here. In the future, we want to replace this stuff anyway with formats and tools that will probably be even more performant. I think this is an important step on the path to replacement because now all of the call sites are obvious.

### Testing

Testing this is a bit of a challenge. The existing logic was not tested or fully documented so the challenge is to ensure that we are producing identical results (this is arguably even more important than producing correct results). Fortunately, our current level of testing offers at least some basic assurances (as you can see by the many failures...). This ensures that everything at least works and produces roughly the same results.

I've also run the pipeline with and without this change and done a detailed comparison of outputs. The outputs actually do slightly differ. In some cases, the spacing is slightly different, and in some cases `None` is replaced with the empty string. I actually think this is a good thing since these are decisions made by the python csv writer and ensure our outputs conform to standards. I compared the contents of these resulting rows and they were identical and parsable.

It's possible in certain edge cases we may encounter more problems, but I still think it is worth it to merge this and monitor for them. It will ultimately make it easier to detect parsing and schema issues.

### DIFF
--- a/short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
@@ -31,7 +31,7 @@ class _TypedDictTSVWriter(DictWriter):
     """
     This is just a convenience class so you don't need to pull out the field names
     """
-    def __init__(self, f: Any, schema: Sequence[Tuple[str]]) -> None:
+    def __init__(self, f: Any, schema: Sequence[Tuple[str, type]]) -> None:
         fieldnames = [field for field, _ in schema]
         super().__init__(f, fieldnames, delimiter="\t")
 
@@ -117,12 +117,6 @@ class _BlastnOutput6ReaderBase(_TypedDictTSVReader):
 
 
 class BlastnOutput6Reader(_BlastnOutput6ReaderBase):
-    """
-    This class is a bit of an oddball due to some compatibility concerns. In addition
-    to the normal parsing stuff it also:
-    1. Ignores comments (lines starting with '#') in tsv files, rapsearch2 adds them
-    2. Supports filtering rows that we consider invalid
-    """
     def __init__(self, f: Iterable[Text], filter_invalid: bool = False, min_alignment_length: int = 0):
         super().__init__(f, _BLASTN_OUTPUT_6_SCHEMA, filter_invalid, min_alignment_length)
 
--- a/short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
@@ -172,7 +172,7 @@ class HitSummaryWriter(_HitSummarySchema, _TypedDictTSVWriter):
 
 
 class _HitSummaryMergedSchema:
-    SCHEMA = [
+    SCHEMA = _HitSummarySchema.SCHEMA + [
         ("contig_id", str),
         ("contig_accession_id", str),
         ("contig_species_taxid", int),
--- a/short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
@@ -1,4 +1,3 @@
-
 from csv import DictReader, DictWriter
 from typing import Any, Iterable, Sequence, Text, Tuple
 
@@ -24,6 +23,8 @@ class _TypedDictTSVReader(DictReader):
         for key, value in row.items():
             if value:
                 row[key] = self._types[key](value)
+            else:
+                row[key] = None
         return row
 
 
@@ -186,3 +187,4 @@ class HitSummaryMergedReader(_TypedDictTSVReader):
 class HitSummaryMergedWriter(_TypedDictTSVWriter):
     def __init__(self, f: Any) -> None:
         super().__init__(f, _HIT_SUMMARY_MERGED_SCHEMA)
+
--- a/short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
@@ -1,12 +1,12 @@
 from csv import DictReader, DictWriter
-from typing import Any, Iterable, Sequence, Text, Tuple
+from typing import Any, Dict, Iterable, Iterator, Sequence, Text, Tuple
 
 # Alignments with e-values greater than 1 are low-quality alignments and associated with
 # a high rate of false-positives. These should be filtered at all alignment steps.
 MAX_EVALUE_THRESHOLD = 1
 
 
-class _TypedDictTSVReader(DictReader):
+class _TypedDictTSVReader(Iterator[Dict[str, Any]]):
     """
     Similar to DictReader but instead of a sequence of fieldnames it takes a sequence of
     tuples, the first element being the field name and the second being the field's type.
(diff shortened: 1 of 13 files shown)


short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
```diff
@@ -0,0 +1,225 @@
...
+        else:
+            super().__init__(tsv_stream)
+
+    @staticmethod
+    def row_is_valid(row, min_alignment_length) -> bool:
+        # GSNAP outputs bogus alignments (non-positive length /
+        # impossible percent identity / NaN e-value) sometimes,
+        # and usually they are not the only assignment, so rather than
+        # killing the job, we just skip them. If we don't filter these
+        # killing the job, we just skip them. If we don't filter these
```
kislyuk, line 163: Duplicate line

kislyuk:
Thanks @morsecodist, this is a great effort and I'm very happy to see the direction we're going with this.

Some architectural questions for this before we proceed (would have been great for a design review, but post-hoc design review is fine too):

- Why use the abc module instead of plain inheritance?
- The csv module has csv.DictReader which is a schematized tabulated format reader. Why not inherit directly from that class? What do you think about having a class for each format, with the schema as a class variable, and __init__ passing that schema to the parent constructor?
- Can you add some unit tests for this? Any time we add a new interface (in this case, the new readers) we should have a unit test for the interface. I recognize the difficulties with integration testing these - but we should at least have unit tests that alert us if this code ceases to correctly handle input that we expect the pipeline to give it, since the pipeline may have a more complex behavior that might mask the issue.

morsecodist:
@kislyuk thanks for the review, good points. The form things are in now is not what I would have chosen if designing from first principles due to a few compatibility concerns, though maybe we should change more. A design review would have been good but a lot of these decisions came up organically as I was going over the code base and wouldn't have been clear beforehand.

- My inheritance structure is more complicated because on `short-read-mngs/idseq-dag/idseq_dag/steps/blast_contigs.py:442` we need a list of the fields from one of the schemas. Previously, I just had the schemas as a class property and the other classes just extended the base class and called it's constructor with the schemas. I intended to keep everything in the class's namespace and to do this I needed access to the schema when using static methods on the class, hence the complicated abc structure. I considered construcing an instance with an empty iterable which would work but it feels hacky. I also considered exporting the schema and using it directly which would probably be cleaner now that I think more about it. I liked the idea of the caller not needing to worry about the notion of schemas but I think I did more harm than good here. I will change it to this unless you have another suggestion.
- I would like to leverage the `DictReader` more (if we added headers we could just use it as is) but I needed to add some custom logic to recognize which variant we were dealing with because it is not clear from the code. I think I could slot this into `DictReader` but with the ammount of custom stuff I felt it wasn't worth it. Now that I've done more testing I probably could figure out which is which at each call site. That would simplify things a lot. I wanted to box how much time I spent here because there is a lot more I want to change but now that I have every usage easily findable I think I can do this pretty quickly so I will set a short time box and see if it is doable.
- I will write some unit tests. The main issue I had is that we have a lot of idseq-dag unit tests that are no longer usable and they don't run with CICD so I felt if I put it in there they would get lost. I tried a bit to see if I could get the unit tests into a workable state but it would be too much for this PR. I will leave a comment over them that they are post-transfer and should be preserved but we should look into this going forward.

I will make the changes I discussed and update soon.

morsecodist:
@kislyuk tests and a few rough edges coming soon but how do you feel about this for a simpler approach? I like it much better than before.

short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
```diff
@@ -0,0 +1,194 @@
...
+_BLASTN_OUTPUT_6_NT_SCHEMA = _BLASTN_OUTPUT_6_SCHEMA + [
+    ("qlen", int),      # query sequence length, helpful for computing qcov
+    ("slen", int),      # subject sequence length, so far unused in IDseq
+]
+
+# This is needed to pass the output fields to blastn on NT (idseq-dag/idseq_dag/steps/blast_contigs.py:441)
+BLASTN_OUTPUT_6_NT_FIELDS = [field for field, _ in _BLASTN_OUTPUT_6_NT_SCHEMA]
+
+# Re-ranked output of blastn.  One row per query.  Two additional columns.
+_BLASTN_OUTPUT_6_NT_RERANKED_SCHEMA = _BLASTN_OUTPUT_6_NT_SCHEMA + [
```
kislyuk, line 67: Consider attaching these directly to the classes:
```
class FooReader(Reader):
    SCHEMA = [
        ("foo", str),
        ("bar", int),
    ]

class BarReader(FooReader):
    SCHEMA = FooReader.SCHEMA + [
        ("xyz", str),
        ("abc", int),
    ]
```
This makes it easier to track and reference them if needed.

short-read-mngs/idseq-dag/tests/unit/util/test_parsing.py
```diff
@@ -0,0 +1,52 @@
...
+                "sstart": 30,
+                "send": 35,
+                "evalue": 0.1,
+                "bitscore": 1.0,
+            })
+
+    def test_writing_error(self):
+        with TemporaryFile("w") as f:
+            writer = BlastnOutput6Writer(f)
+            self.assertRaises(Exception, lambda : writer.writerow({"bad key": 1}))
```
kislyuk, line 38: Consider using a context manager for readability:
```
with self.assertRaises(Exception):
    writer.writerow(...)
```

kislyuk (approved):

kislyuk:
I can't say that I've checked all the logic updates - there's a lot - but I trust you've kept good track of everything and it's a definite improvement.

I would prefer to see unit test coverage with a good and bad row for each reader class, but I won't insist on adding it all now.

short-read-mngs/idseq-dag/idseq_dag/util/parsing.py
morsecodist: I see the value in this but because we have both readers and writers do this I would need to either:
- Duplicate the schemas between the reader and writer, which seems messy to me
- Add the schema to a common base class between the reader and the writer, which seems like what I have now but with more boilerplate
- Reference the schema from the reader in the writer or visa versa. This is probably the strongest option in my opinion and doesn't have a huge downside. Subjectively both using a common schema seems more sensible to me than one using the other's schema but if you feel strongly I could go with this.

kislyuk: Duplicating would be an issue of course 😄 

Attaching the data directly to the class is more readable and self-explanatory, and less likely to cause issues later.

I agree that both your second option (mix-ins) and third option (class variable references across class declarations) are good. Whichever one results in fewer lines of code seems best to me 😄

morsecodist:
@kislyuk added the tests, just going to do one more branch test for sanity then merge.

PR #71, opened 2021-01-12
enable the ability to create generic consensus genomes

katrinakalantar:
### Description

This PR makes it possible to create a generic consensus genome using the wdl workflow by passing the appropriate inputs (reference genome and primer files) and making the `filterReads` step optional. To generate generic consensus genomes, you would now need to include `"filter_reads": false` in the .json of wdl inputs. 

note: this PR does not add additional error handling logic to ensure that a reference genome is of the correct format (.fasta) or any additional error handling.

### Tests

- Tested locally on several different viruses and verified that the results look appropriate via comparison to ground truth genomes.
- Tested on staging using the `run_sfn.py` script. In particular, I tested that 1. it works with a "generic" non-sars-cov-2 reference genome (tested CHKV genome) and 2. verified that it still generates the same results for two different sars-cov-2 samples - one with PE inputs using ARTIC v3 and one with SE inputs using MSSPE.

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -1,5 +1,5 @@
 # The following pipeline was initially based on previous work at: https://github.com/czbiohub/sc2-illumina-pipeline
-# workflow version: consensus-genomes-1.5.2
+# workflow version: consensus-genomes-1.5.0
 version 1.0
 
 workflow consensus_genome {
@@ -58,7 +58,7 @@ workflow consensus_genome {
             docker_image_id = docker_image_id
     }
 
-    if (filter_reads){
+    if (filter_reads) {
         call FilterReads {
             input:
                 prefix = prefix,
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -25,6 +25,7 @@ workflow consensus_genome {
 
         # all of the following seem to be parameters in params.* of the nextflow pipeline;
         # could they be extracted to a separate params file? put in .json input to wdl?
+        Boolean filter_reads = true
         Boolean trim_adapters = true
 
         Float ivarFreqThreshold = 0.9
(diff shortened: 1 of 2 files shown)


tfrcarvalho (approved):

consensus-genome/run.wdl
```diff
@@ -25,6 +25,7 @@ workflow consensus_genome {
 
         # all of the following seem to be parameters in params.* of the nextflow pipeline;
         # could they be extracted to a separate params file? put in .json input to wdl?
+        Boolean filter_reads = true
```
jshoe, line 28: Could you bump the `# workflow version: consensus-genomes-1.4.2` above to 1.5.2 ?  (Maybe we don't need that comment but it seems like we've been using it so far)

consensus-genome/run.wdl
```diff
@@ -1,5 +1,5 @@
 # The following pipeline was initially based on previous work at: https://github.com/czbiohub/sc2-illumina-pipeline
-# workflow version: consensus-genomes-1.4.2
```
tfrcarvalho, line 2: IT would be `1.5.0` then

jshoe: Oops, yes :)

consensus-genome/run.wdl
```diff
@@ -57,19 +58,21 @@ workflow consensus_genome {
...
     }
 
-    call FilterReads {
-        input:
-            prefix = prefix,
-            fastqs = RemoveHost.host_removed_fastqs,
-            ref_fasta = ref_fasta,
-            kraken2_db_tar_gz = kraken2_db_tar_gz,
-            docker_image_id = docker_image_id
+    if (filter_reads){
```
morsecodist, line 61: nitpick: missing space

morsecodist (approved):
Looks good

consensus-genome/run.wdl
katrinakalantar: thank you! updated to the correct version now.

jshoe:
Screen Shot 2021-01-20 at 1 49 50 PM

Just FYI @katrinakalantar, if you wanted to consolidate the "committer" appearance, you can run something like this locally so that the email is recognized by Github:

```
git config --global user.name "John Doe"
git config --global user.email johndoe@example.com
```

PR #73, opened 2021-01-14
Filter taxids out of hit summaries

morsecodist:
This actually contains two changes, a bit confusing but one won't work without the other.

First, I make taxids all strings. We have a big problem in the codebase of taxids being strings/numbers. For filtering, they must be strings. In some places they had to be ints because we do math with them and compare them as numbers sometimes. I think a cleaner approach overall would actually be to make them always ints but it is much safer to go in the string direction since this is explicitly used in way more places and we very rarely do math on the ids.

Second, I filter out taxids from the hit summary from alignment. This has the effect of filtering them out of the refined hits ummary as well. It took a lot of analyzing the code to come to this conclusion but I am pretty confident about it at this point. I added an explainatory comment. In general no taxids will be in the refined hit summary if they are not in the hit summary.

### DIFF
--- a/short-read-mngs/idseq-dag/idseq_dag/util/lineage.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/util/lineage.py
@@ -1,10 +1,10 @@
-from typing import List
+from typing import List, Sequence, Iterator
 
 INVALID_CALL_BASE_ID = -(10**8)
-NULL_SPECIES_ID = -100
-NULL_GENUS_ID = -200
-NULL_FAMILY_ID = -300
-NULL_LINEAGE = (str(NULL_SPECIES_ID), str(NULL_GENUS_ID), str(NULL_FAMILY_ID))
+NULL_SPECIES_ID = "-100"
+NULL_GENUS_ID = "-200"
+NULL_FAMILY_ID = "-300"
+NULL_LINEAGE = (NULL_SPECIES_ID, NULL_GENUS_ID, NULL_FAMILY_ID)
 
 DEFAULT_BLACKLIST_S3 = 's3://idseq-public-references/taxonomy/2018-04-01-utc-1522569777-unixtime__2018-04-04-utc-1522862260-unixtime/taxon_blacklist.txt'
 DEFAULT_WHITELIST_S3 = 's3://idseq-public-references/taxonomy/2020-02-10/respiratory_taxon_whitelist.txt'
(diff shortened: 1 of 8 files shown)


tfrcarvalho (commented):
My memory of the original issue is a bit fuzzy, but it does look like you address the original issue of filtering out of `output_hitsummary`.
A couple of comments:
* I understand but am not fond of making taxIds into string instead of guaranteeing they are ints. Are we in any way exposed to bad taxids? Do we need to do any validation before accepting input as strings?
* How did you test it? Did you go over the edge cases in the ticket?

In general,  it would be good to include a testing section describing how the solution was tested.
nit: When there is a mapping to a clubhouse ticket, it would be great to include it in the title and/or description.

short-read-mngs/idseq-dag/idseq_dag/util/lineage.py
```diff
@@ -58,33 +54,33 @@ def fill_missing_calls(cleaned_lineage: List[str]) -> List[str]:
...
-        tax_level -= 1
     return result
 
 
-def blank(taxid_int):
+def _blank(taxid_int: int):
     return 0 > taxid_int > INVALID_CALL_BASE_ID
 
 
 def fill_missing_calls_tests():
```
tfrcarvalho, line 70: Do you know the history behind this? Is it still being used? Could not find it

morsecodist: I did a global search and I could not. I made as many of the functions in this module private as I could to avoid confusion.

morsecodist:
@tfrcarvalho 

> Are we in any way exposed to bad taxids?

They are strings in the lookup dicts but they shouldn't be bad anywhere else. I agree the strings are a bit of a hack. I'd prefer to go ints but I felt it was a bit riskier. Do you think I should go ints?

Testing wrietup incoming

short-read-mngs/idseq-dag/idseq_dag/util/lineage.py
tfrcarvalho: Not worth deleting?

tfrcarvalho:
> They are strings in the lookup dicts but they shouldn't be bad anywhere else. I agree the strings are a bit of a hack. I'd prefer to go ints but I felt it was a bit riskier. Do you think I should go ints?

You have a better understanding of the immediate effort/risk (if our inputs do not have the chance of having malformed tax ids or the parser check the format , there is not much of a risk)... so, will leave final decision up to you.

PR #75, opened 2021-01-21
increase sj collapsed limit

morsecodist:
We got a crash from this parameter so I doubled it.

### DIFF
--- a/short-read-mngs/idseq-dag/idseq_dag/steps/run_star.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/steps/run_star.py
@@ -271,6 +271,9 @@ class PipelineStepRunStar(PipelineStep):
         #  when we update the rest of our rng seeds.
         rng_seed = '777'
 
+        # default is 1000000, this caused a crash so it was doubled
+        limit_out_sj_collapsed = '2000000'
+
         cpus = str(multiprocessing.cpu_count())
         cd = output_dir
         cmd = 'STARlong' if use_starlong else 'STAR'
@@ -281,6 +284,7 @@ class PipelineStepRunStar(PipelineStep):
             '--outReadsUnmapped', 'Fastx',
             '--outFilterMismatchNmax', '999',
             '--clip3pNbases', '0',
+            '--limitOutSJcollapsed', limit_out_sj_collapsed,
             '--runThreadN', cpus,
             '--runRNGseed', rng_seed,
             '--genomeDir', genome_dir,


short-read-mngs/idseq-dag/idseq_dag/steps/run_star.py
```diff
@@ -271,6 +271,9 @@ def run_star_part(self,
         #  when we update the rest of our rng seeds.
         rng_seed = '777'
 
+        # default is 1000000, this caused a crash so it was doubled
```
tfrcarvalho, line 274: Do we know what the parameter refers to?

morsecodist: `max number of collapsed junctions`

tfrcarvalho: ok... i guess does not had that much info without understanding the tool

tfrcarvalho (approved):

short-read-mngs/idseq-dag/idseq_dag/steps/run_star.py
morsecodist: From the manual, if you want more context: 

```
SJ.out.tabcontains high confidence collapsed splice junctions in tab-delimited format.  Note thatSTAR  defines  the  junction  start/end  as  intronic  bases,  while  many  other  software  define  them  asexonic bases.  The columns have the following meaning

column 1:  chromosome

column2:  first base of the intron (1-based)

column 3:  last base of the intron (1-based)

column 4:  strand (0:  undefined, 1:  +, 2:  -)

column 5:  intron  motif:  0:  non-canonical;  1:  GT/AG,  2:  CT/AC,  3:  GC/AG,  4:  CT/GC,  5:AT/AC, 6:  GT/AT

column 6:  0:  unannotated, 1:  annotated (only if splice junctions database is used)

column 7:  number of uniquely mapping reads crossing the junction

column 8:  number of multi-mapping reads crossing the junction

column 9:  maximum spliced alignment overhang

The filtering for this output file is controlled by the --outSJfilter parameters,  as described in Section 14.16.  Output Filtering:  Splice Junctions.
```

tfrcarvalho: I was looking for just a more informative comment to add to the code, for the next (maybe on-call) person to do a quick decision of keeping to increase or just accept the error...

PR #76, opened 2021-01-27
added host genome generation wdl

morsecodist:
I created a basic wdl file for generating host genomes. This will be deployed automatically alongside the `short-read-mngs` wdl files. Developers can use the single wdl infrastructure to manually fill in a JSON template with the values it needs. This may not be the cleanest solution but it gets us off of running idseq-dag in one of the last few places and it's simple. I also think it makes sense to keep here because host genome generation is part of the `short-read-mngs` pipeline. 

I am testing this now by running it to generate the new host genomes. I will update the wiki if it works.

### DIFF
--- a/short-read-mngs/host_genome_generation.wdl
+++ b/short-read-mngs/host_genome_generation.wdl
@@ -46,7 +46,7 @@ workflow idseq_host_filter {
     input:
       docker_image_id = docker_image_id,
       s3_wd_uri = s3_wd_uri,
-      host_name = host_name
+      host_name = host_name,
       host_fasta = host_fasta,
       ercc_fasta = ercc_fasta,
       ercc_gtf = ercc_gtf,
--- a/short-read-mngs/host_genome_generation.wdl
+++ b/short-read-mngs/host_genome_generation.wdl
@@ -19,7 +19,7 @@ task RunGenerateHostGenome {
     --output-files '["fasta_with_ercc.fa", "gtf_with_ercc.fa", "~{host_name}_STAR_genome.tar", "~{host_name}_bowtie2_genome.tar"]' \
     --output-dir-s3 '~{s3_wd_uri}' \
     --additional-files '{ "ercc_fasta": "~{ercc_fasta}", "ercc_gtf": "~{ercc_gtf}" }' \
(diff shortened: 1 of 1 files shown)


short-read-mngs/host_genome_generation.wdl
```diff
@@ -0,0 +1,61 @@
+version 1.0
```
tfrcarvalho, line 1: Should this be in short-read-mngs?

short-read-mngs/host_genome_generation.wdl
```diff
@@ -0,0 +1,61 @@
...
+  set -euxo pipefail
+  idseq-dag-run-step --workflow-name host_genome_generation \
+    --step-module idseq_dag.steps.generate_host_genome \
+    --step-class PipelineStepGenerateHostGenome \
+    --step-name generate_host_genome \
+    --input-files '[["~{host_fasta}"]]' \
+    --output-files '["fasta_with_ercc.fa", "gtf_with_ercc.fa", "~{host_name}_STAR_genome.tar", "~{host_name}_bowtie2_genome.tar"]' \
+    --output-dir-s3 '~{s3_wd_uri}' \
+    --additional-files '{ "ercc_fasta": "~{ercc_fasta}", "ercc_gtf": "~{ercc_gtf}" }' \
+    --additional-attributes '{ "host_name": "~{host_name}", "max_star_part_size": null }'
```
tfrcarvalho, line 22: Is the null `max_star_part_size ` parameter required?

tfrcarvalho (approved):
Couple of small notes.

short-read-mngs/host_genome_generation.wdl
morsecodist: I think so since it is part of that pipeline, just run by us far less often. I felt it was less confusing to keep them together, though if you feel it's better to have it's own top level directory I can move.

short-read-mngs/host_genome_generation.wdl
morsecodist: It isn't, removing.

morsecodist:
Went with a different approach

PR #78, opened 2021-01-30
Add benchmark truth files to manifest

kislyuk:
Also:

- the atcc fastq files seem to have been switched up with the test titles
- fix typo


katrinakalantar (approved):
The raw .fastq to truth file mappings look good to me!

PR #79, opened 2021-02-04
Add utility functions to read truth files, compute AUPR

kislyuk:


tests/short-read-mngs/auto_benchmark/harvest.py
```diff
@@ -142,7 +147,76 @@ def harvest_sample(sample, outputs_json, taxadb):
...
+    # Using raw abundances as proxies for confidence score per Ye2009 methodology
+    # https://www.cell.com/cell/fulltext/S0092-8674(19)30775-5#fig2
+    nt_confidence_scores = [i["reads_dedup"] / total_reads for i in ans["taxa"]["NT"].values()]
+    # TODO: what's the correct placeholder value here? Setting it to 0 causes AUPR to blow out to 0.
+    nt_confidence_scores += [1e-100] * len(nt_missed_taxa)
+    ans["aupr"] = adjusted_aupr(nt_correctness_labels, nt_confidence_scores, force_monotonic=False)["aupr"]
+
+    truth_sum = sum(ans["truth"].values())
+    relative_abundances_diff = [
+        ans["truth"][taxon]/truth_sum - ans["taxa"]["NT"].get(taxon, {}).get("reads_dedup", 1e-100)/total_reads
```
kislyuk, line 215: Surfaced in conversation with @mlin: total_reads needs to be recalculated for only truth taxa

tests/short-read-mngs/auto_benchmark/harvest.py
```diff
@@ -142,7 +147,76 @@ def harvest_sample(sample, outputs_json, taxadb):
...
+        tax_id for tax_id in ans["truth"]
+        if tax_id not in ans["taxa"]["NT"]
+    ]
+    nt_correctness_labels = [
+        1 if tax_id in ans["truth"] else 0
+        for tax_id in ans["taxa"]["NT"]
+    ]
+    nt_correctness_labels += [1] * len(nt_missed_taxa)
+    total_reads = sum([i["reads_dedup"] for i in ans["taxa"]["NT"].values()])
+    # Using raw abundances as proxies for confidence score per Ye2009 methodology
```
kislyuk, line 206: typo: s/2009/2019/

kislyuk:
We will also need a strategy for combining read counts from nr and nt for the purpose of abundance estimate reporting (so far in this PR I have only used nt counts). @katrinakalantar could you advise on the best way to do this?

katrinakalantar:
Good question. Here are a few general thoughts on this. I can see a few ways of doing this, each with pros/cons...
1. Compute metrics for NT and NR separately - given that algorithmic changes specific to the alignment databases will only affect single metrics, there may be utility to track the metrics separately. Of course, this doesn't reflect the value of combining the metrics.
2. Apply a minimal threshold filter to require concordance on NT and NR (specifically, require NT > 0 and NR > 0 for any particular taxa). Then, use the NT counts. This is an approach that many users take when actually interpreting the data in the IDseq reports to improve specificity. However, it does not enable detection of divergent taxa where counts on NR alone may be of interest. It also does not reflect "raw results". 
3. Apply the algorithm for combining NT/NR that was implemented here: https://github.com/chanzuckerberg/idseq-workflows/pull/51. This is the most rigorous method of combining NT and NR because it gives one taxID to each read. However, it is under active refinement to improve performance for a small number of divergent viral edge cases and is not currently used for presentation of production results.

Another more naive approach that I'm not certain about would be to sum together the NT and NR counts - unfortunately this means that single reads would potentially be counted twice. This does allow you to inherit the sensitivity from each, but will take a hit on perceived specificity.

mlin:
Built on this in #83

PR #80, opened 2021-02-06
Migrate status JSON files out of idseq-dag

mlin:
Delete the code for creating these JSON files (to be read by webapp to display pipeline status) from idseq-dag, in favor of posting them from a miniwdl plugin.

The status JSON includes a markdown description for each step, which in some cases idseq-dag customized dynamically to reflect details of the specific inputs. We keep that idseq-dag logic and transmit the description through a WDL String output from each task to be read by the plugin.

### DIFF
--- a/short-read-mngs/experimental.wdl
+++ b/short-read-mngs/experimental.wdl
@@ -22,6 +22,7 @@ task GenerateTaxidFasta {
     --additional-attributes '{}'
   >>>
   output {
+    String step_description_md = read_string("taxid_fasta_out.description.md")
     File taxid_annot_fasta = "taxid_annot.fasta"
     File? output_read_count = "taxid_fasta_out.count"
   }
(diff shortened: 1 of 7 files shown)


morsecodist (approved):
The changes in here look good but what is the plan to move to the markdown files on the webapp? We can't deploy this until that is handled.

mlin:
Good point @morsecodist -- per @tfrcarvalho suggestion, I restored the existing code to update status.json within idseq-dag, and changed the plugin to create status2.json instead. This will make a smoother transition, we can re-delete the idseq-dag bits once the webapp is confirmed working with status2.json. (The diff still deletes some invocations of `update_status_json_file()` that are in dead code anyway.)

kislyuk:
Just noting down a follow-on thought that we've discussed a few times: it would be great to minimize the number of tasks (pipeline steps) that have dynamic descriptions, and for those that have static descriptions, write those down as markdown comments in the WDL file itself, so the frontend can consume them from the static copy. (Building on that, for dynamic descriptions, we could use a templating convention where the docstring uses a jinja or ERB template with values that can be filled in from task inputs or outputs.)

mlin:
They certainly were an inconvenience! One limitation is that WDL doesn't yet have multi-line string literals (task commands excepted), so it's inconvenient to put a lengthy markdown in them. That's been discussed in the standards forum, but hasn't made its way through yet.

PR #84, opened 2021-03-05
Update CG Dockerfile

kislyuk:
This makes multiple updates to the Consensus Genomes Dockerfile:
- Remove conda and replace packages installed via conda with packages installed via apt and pip3 if not available in apt
- Install ivar
- Repackage VADR for deterministic installation
- Install VADR

While this image should be suitable for existing pipeline functions, I'm still working on the taxoniq integration (PR coming up).

In addition, I have skipped the installation of VADR models (see https://github.com/ncbi/vadr#models). I think we should treat them as we treat other reference data (store on S3, pass by reference as an input parameter, and decouple from the package codebase) but happy to hear other opinions. We also need to determine which models to use. Do we only intend to use VADR on coronavirus samples, or are we shooting for more general application? @katrinakalantar can you confirm which VADR models we should use?


consensus-genome/Dockerfile
```diff
@@ -1,30 +1,88 @@
...
 
 LABEL maintainer="IDseq Team idseq-tech@chanzuckerberg.com"
 LABEL description = "Image for consensus genome by metagenomic sequencing with spiked primer enrichment or amplicon sequencing"
 
+RUN sed -i s/archive.ubuntu.com/us-west-2.ec2.archive.ubuntu.com/ /etc/apt/sources.list; \
+  echo 'APT::Install-Recommends "false";' > /etc/apt/apt.conf.d/98idseq; \
+  echo 'APT::Install-Suggests "false";' > /etc/apt/apt.conf.d/99idseq
 
-RUN apt-get -qq update && apt-get -qq -y install curl locales zip \
+RUN apt-get -qq update && apt-get -qq -y install \
```
tfrcarvalho, line 11: We are not freezing package versions?

tfrcarvalho (approved):
How did the image size turned out?

consensus-genome/Dockerfile
kislyuk: APT package versions are stable for each version of Ubuntu. As a best practice, we should only be using Ubuntu LTS versions (16.04, 18.04, 20.04, ...). We only need to worry about changes when we change the Ubuntu version (as we do here, along with several other changes, so the whole pipeline will need to be revalidated - but that was going to have to happen anyway with the nanopore and general CG changes).

kislyuk:
The image size went down from 1.8GB to 1.1GB.

consensus-genome/Dockerfile
kislyuk: (I also looked at how the package versions changed from conda to APT, and all of the upstream version changes are minor, but semantically there may still be differences that we will need to watch out for.)

jshoe (approved):
LGTM

consensus-genome/Dockerfile
```diff
@@ -1,30 +1,88 @@
...
+  seqtk \
+  bedtools \
+  build-essential \
+  autoconf \
   && locale-gen en_US.UTF-8
 
-# Install miniconda and configure it
-ENV MINICONDA_SHA256 bb2e3cedd2e78a8bb6872ab3ab5b1266a90f8c7004a22d8dc2ea5effeb6a439a
+RUN curl -O -L http://ftp.osuosl.org/pub/ubuntu/pool/universe/i/ivar/ivar_1.2.2+dfsg-3_amd64.deb && \
+  dpkg -i ivar_1.2.2+dfsg-3_amd64.deb && \
```
katrinakalantar, line 51: If possible, it would be great to wrap the update to iVar v 1.3.1 into this update. We need access to the `-x` parameter that was added in v1.3.1 here: https://github.com/andersen-lab/ivar/releases to accommodate for trimming in the SNAP protocol.

katrinakalantar (commented):
Added a comment about iVar version. Otherwise, looks good to me!

katrinakalantar:
Regarding `which VADR models we should use`, I think it would be ideal to start with VADR for sars-cov-2 consensus genomes only. In which case we can limit the scope of the reference files to only the sars-cov-2 release. Once we have a sense of the broader range of usage for the generalized cg workflow we could expand this.

kislyuk:
@katrinakalantar I have followed your guidance to update iVar to 1.3.1 (building it from source instead of using an Ubuntu package; unlike the other addition, building iVar from source was uneventful). I also loaded the VADR Coronavirus reference model and deposited it at s3://idseq-public-references/consensus-genome/vadr-models-corona-1.1.3-1.tar.gz. I started a reference provenance list at https://github.com/chanzuckerberg/idseq-workflows/blob/akislyuk-cg-updates/consensus-genome/README.md#reference-files as we discussed.

consensus-genome/Dockerfile
katrinakalantar: If we are not freezing version here, do we still have a way to document / make it visible to the user which versions of these end up being installed?

PR #87, opened 2021-03-18
Generalized CG updates

kislyuk:
- Added https://github.com/chanzuckerberg/taxoniq to Dockerfile, implemented new FetchSequenceByAccessionId task and unit test for it
- Updated WDL grammar to 1.1 to make null input checking easier
- Added a positive and negative workflow unit test for CG with mumps virus subsample inputs provided by @katrinakalantar 

### Notes
- More unit testing is desirable with different input conditions
- The Docker image size has grown substantially; updating taxoniq index SOP TBD
- TODO: It would be best to raise an error if both ref_fasta and ref_accession_id is set, for the avoidance of confusion.

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -16,10 +16,12 @@ workflow consensus_genome {
         File ercc_fasta = "s3://idseq-public-references/consensus-genome/ercc_sequences.fasta"
         File kraken2_db_tar_gz  # TODO: make this optional; only required if filter_reads == true, even for Illumina
         File primer_bed = "s3://idseq-public-references/consensus-genome/artic_v3_primers.bed" # Only required for Illumina
-        File? ref_fasta # Only required for Illumina
-        String? ref_accession_id # Only required for Illumina
+
+        File? ref_fasta # Only required for Illumina (ONT SC2 reference is built into ARTIC); takes precedence over ref_accession_id
+        String? ref_accession_id # Only required for Illumina; has no effect if ref_fasta is set
+
         File ref_host
-        String technology # Input sequencing technology ("Illumina" or "ONT")
+        String technology # Input sequencing technology ("Illumina" or "ONT"); ONT only works with SC2 samples (SC2 reference is built into ARTIC)
 
         # Sample name: include in tags and files
         String sample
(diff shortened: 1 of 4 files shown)


consensus-genome/run.wdl
```diff
@@ -16,7 +16,8 @@ workflow consensus_genome {
         File ercc_fasta = "s3://idseq-public-references/consensus-genome/ercc_sequences.fasta"
         File kraken2_db_tar_gz  # TODO: make this optional; only required if filter_reads == true, even for Illumina
         File primer_bed = "s3://idseq-public-references/consensus-genome/artic_v3_primers.bed" # Only required for Illumina
-        File ref_fasta # Only required for Illumina
+        File? ref_fasta # Only required for Illumina
```
katrinakalantar, line 19: Would it be useful to add a comment here or other documentation to specify which of these is the primary reference when both `ref_fasta` and `ref_accession_id` are selected? below, it seems we take `ref_fasta` as the primary if supplied.

consensus-genome/run.wdl
```diff
@@ -313,7 +321,25 @@ task ValidateInput{
...
+}
+
+task FetchSequenceByAccessionId {
+    input {
+        String accession_id
+        String docker_image_id
+    }
 
+    command <<<
+    taxoniq get_from_s3 --accession-id "~{accession_id}" > sequence.fa
```
katrinakalantar, line 333: nit: indent here?

consensus-genome/test/test_wdl.py
```diff
@@ -119,19 +123,61 @@ def test_sars_cov2_ont_cg(self):
             for filename in output:
                 self.assertGreater(os.path.getsize(filename), 0)
 
-    # note: attempted unit test to test that .fq.gz inputs are correctly 
-    #       converted to .fastq.gz formats to ensure they are read by guppyplex 
-    #       in ApplyLengthFilter 
     def test_sars_cov2_ont_cg_input_file_format(self):
+        """
```
katrinakalantar, line 127: Thank you for modifying that comment I had left to the more appropriate structure!

katrinakalantar (commented):
This looks good to me! Made a few minor comments.

katrinakalantar (approved):

katrinakalantar (approved):
Looks good to me!

consensus-genome/run.wdl
kislyuk: Great suggestion, added comment.

consensus-genome/run.wdl
kislyuk: Indented. The WDL IDE plugin I'm using doesn't want to indent it, but I'm thinking that's probably a bug in the plugin, not a WDL convention.

PR #88, opened 2021-03-18
simplify pytest fixture scheme

mlin:
* Consolidate helper functions into `test_util/` module (instead of gratuitously packaging them as fixtures)
* Move unit tests from `tests/short-read-mngs` to `short-read-mngs/test`
  * (`tests/short-read-mngs/auto_benchmark` to be moved at a later date, to avoid messy merge with other wip on that)


short-read-mngs/test/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+The [pytest](https://docs.pytest.org/en/stable/) suite has two stages:
+
+1. The workflow runs end-to-end on one or more test inputs.
+2. Task-level unit tests exercise error paths or corner cases that aren't otherwise covered by the workflow tests.
+
+The workflow tests are triggered by session-scoped pytest fixtures, making their results available to the task-level cases. This two-stage structure allows the task unit tests to reuse intermediate inputs/outputs generated by the workflow tests, minimizing the boilerplate needed to express them.
+
+For example, `./conftest.py` has the workflow test fixture for a ssynthetic FASTQ pair to run using viral databases, `short_read_mngs_bench3_viral_outputs`. When pytest begins a task test case such as `host_filter/test_RunValidateInput.py`, the fixture triggers it to first run the end-to-end workflow *if it hasn't already been* in the current session. The test case then re-runs just one step of the workflow, using inputs modified from those generated in the end-to-end run. Many different test cases can similarly base themselves on individual steps in the workflow run triggered by the `short_read_mngs_bench3_viral_outputs` fixture.
```
kislyuk, line 10: Typo: ssynthetic

kislyuk, line 10: > When pytest begins a task test case such as `host_filter/test_RunValidateInput.py`, the fixture triggers it to first run the end-to-end workflow *if it hasn't already been run* in the current session.

Just as in the code itself, I'm having trouble following the execution flow (what triggers what) in this sentence. Perhaps it would be clearer to say:

> When pytest finds and starts a task test case such as `host_filter/test_RunValidateInput.py`, it first runs the [conftest.py fixtures](https://docs.pytest.org/en/stable/writing_plugins.html#conftest-py-plugins), including `short_read_mngs_bench3_viral_outputs` - which first runs the end-to-end workflow *if it hasn't already been run* in the current session.

short-read-mngs/test/README.md
```diff
@@ -0,0 +1,30 @@
+# short-read-mngs unit tests
+
+The [pytest](https://docs.pytest.org/en/stable/) suite has two stages:
+
+1. The workflow runs end-to-end on one or more test inputs.
+2. Task-level unit tests exercise error paths or corner cases that aren't otherwise covered by the workflow tests.
+
+The workflow tests are triggered by session-scoped pytest fixtures, making their results available to the task-level cases. This two-stage structure allows the task unit tests to reuse intermediate inputs/outputs generated by the workflow tests, minimizing the boilerplate needed to express them.
```
kislyuk, line 8: This is a great introduction, and it's made more readable by the formatting update. From a naive perspective, there's just one question that it does not answer: what precisely triggers the fixtures' execution? The best resource I've been able to find is https://docs.pytest.org/en/stable/writing_plugins.html#conftest-py-plugins - consider linking it here.

kislyuk (approved):
This is a great improvement! Makes it clearer what's going on, and it should become even easier to see as we add more test cases.

short-read-mngs/test/README.md
mlin: Thanks, good point. conftest.py is loaded implicitly, but the fixtures aren't actually executed until starting a test case that actually uses them. A test case declares it needs a fixture by having an argument (to the test function) with the same name as a known fixture. That's a little too clever for its own good, we probably agree, but it's convenient once the trick is revealed.

PR #89, opened 2021-03-22
Updates to address gaps identified in initial ont validation

katrinakalantar:
**Summary**
During validation of the ONT nanopore pipeline, there were a few issues that were identified. This PR resolves those issues as well as modifying some of the default parameters to ensure quality of genomes generated by the ONT nanopore pipeline matches the quality of existing genomes submitted to public data repositories.

*Issues Addressed*
- [x] input raw data for the ONT pipeline must end in .fastq / .fastq.gz to be recognized by the `artic guppyplex` command in `ApplyLengthFilter` (i.e. will not accept samplename.fq or samplename.fq.gz)
- [x] VADR sometimes fails in the `RunVadr` step (for both Illumina and ONT technologies). The error is as follows...
`ERROR in cmalign_run(), cmalign failed in a bad way, see vadr-output/vadr-output.vadr.NC_045512.align.r1.s0.stdout for error output`. This occurs after generating the necessary outputs, so added functionality to handle this error.

*Parameters Updated*
- [x] `normalise` parameter: genomes are higher-quality with `normalise = 1000` than the initially-specified default of `normalise = 200`
- [x] `medaka_model` parameter: medaka model `r941_min_high_g360` has produced better results (matching the published version of associated genomes) than the previously specified medaka model. Additionally this value matches the defaul model implemented by medaka library itself.
- [x] `min_length` parameter: the min_length value in the `ApplyLengthFilter` function results in most of the tested clearlabs data being filtered out. The `min_length` value has been reduced to match the clearlabs value of 350. The output genome remains high-quality.

**Integration Testing**
After these updates were implemented, seven ONT Artic v3 datasets obtained from SRA were run through the ONT pipeline and benchmarked against their final GISAID genomes. The results obtained by this pipeline were identical to those uploaded to GISAID (with the exception of a few samples where uploaded genomes contained 2 Ns that were now called as reference), all samples passed VADR checks and looked good when QC-checked in NextClade.


consensus-genome/test/test_wdl.py
```diff
@@ -100,6 +100,19 @@ def test_sars_cov2_ont_cg(self):
             for filename in output:
                 self.assertGreater(os.path.getsize(filename), 0)
 
+    # note: attempted unit test to test that .fq.gz inputs are correctly 
+    #       converted to .fastq.gz formats to ensure they are read by guppyplex 
+    #       in ApplyLengthFilter 
+    def test_sars_cov2_ont_cg_input_file_format(self):
+        fastqs_0 = os.path.join(os.path.dirname(__file__), "Ct20K.fq.gz")
+        args = ["sample=test_sample", f"fastqs_0={fastqs_0}", "technology=ONT"]
+        res = self.run_miniwdl(args)
```
kislyuk, line 109: I suggest scoping down this test to exercise just the task with the relevant logic, using the `task="ValidateInput"` kwarg here as discussed.

kislyuk (approved):

kislyuk:
Approved pending minor changes.

consensus-genome/run.wdl
```diff
@@ -267,16 +272,43 @@ task ValidateInput{
...
             exit 1
         fi
+        # ONT guppyplex requires files are in .fastq format with .fastq extension (not .fq)
+        FILE=~{fastqs[0]}
+        #if [[ `file $FILE | grep compressed | wc -l` == 1 ]]; then
+        if [[ "${FILE##*.}" == "gz" ]]; then
+            gunzip $FILE
+            FILE="${FILE%.*}"
+        fi
+        cp $FILE "~{prefix}validated.fastq"
```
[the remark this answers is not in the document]
katrinakalantar: Thanks! As we discussed, I had to revert this back to using `cp` due to errors with `mv` that happen as a function of the way we containerize each task.

consensus-genome/test/test_wdl.py
katrinakalantar: Good call - I have updated the test to only run the `ValidateInput` task and verify that the outputs files are of the format `.fastq.gz` with file size > 0.

PR #90, opened 2021-03-24
Update ivar trim to address issues with trimming in SWIFT/SNAP protocol

katrinakalantar:
A fix was recently implemented in iVar v. 1.3.1 that addresses an issue with trimming that was identified for some amplicon-based sequencing protocols (e.g. SWIFT, SNAP), where polymerases can add additional bases and lead mismatches.

Prior to this change, the Dockerfile was updated to include this latest version of iVar (https://github.com/chanzuckerberg/idseq-workflows/pull/84). This PR updates the `ivar trim` call in the `TrimPrimers` step to leverage the recent change. In particular, it makes the `ivar trim` command dependent on the primer trimming .bed file. For the SNAP protocol (`snap_primers.bed`) we now implement a 5-bp offset, consistent with recommendation from the developers of iVar. For all remaining supported primer sets (not polymerase-based) we do not use the primer offset.

This PR also includes one additional test to verify the parameter's impact on downstream depth of coverage for a test SNAP dataset.  

**Testing**
- Tested that when running the workflow with a SNAP dataset and the SNAP .bed file, the primers are trimmed with 5-bp buffer (verified by viewing the primertrimmed.bam file)
- Tested that when running the workflow with a non-SNAP (Artic) dataset with the artic .bed file, the primers are trimmed at their true locations without any buffer (verified by viewing the primertrimmed.bam file)


consensus-genome/run.wdl
```diff
@@ -545,7 +545,13 @@ task TrimPrimers {
 
         samtools view -F4 -q "~{samQualThreshold}" -o ivar.bam "~{alignments}"
         samtools index ivar.bam
-        ivar trim -e -i ivar.bam -b "~{primer_bed}" -p ivar.out
+        #The SNAP protocol results in primer position offsets due to polymerase...if the data is trimming
```
jshoe, line 548: nit: extra space after the `#` ?

consensus-genome/run.wdl
```diff
@@ -545,7 +545,13 @@ task TrimPrimers {
...
+        #SNAP primers, add offset allowance of 5 bp in accordance with ivar developer recommendation
+        if [[ `basename "~{primer_bed}"` == "snap_primers.bed" ]]; then
+            ivar trim -x 5 -e -i ivar.bam -b "~{primer_bed}" -p ivar.out
```
jshoe, line 551: nit: could have 5 / 0 in a variable, might matter more if the command gets longer

jshoe (commented):
Looks good to me! I will leave approval to @kislyuk

consensus-genome/run.wdl
```diff
@@ -545,7 +545,13 @@ task TrimPrimers {
 
         samtools view -F4 -q "~{samQualThreshold}" -o ivar.bam "~{alignments}"
         samtools index ivar.bam
-        ivar trim -e -i ivar.bam -b "~{primer_bed}" -p ivar.out
+        #The SNAP protocol results in primer position offsets due to polymerase...if the data is trimming
+        #SNAP primers, add offset allowance of 5 bp in accordance with ivar developer recommendation
+        if [[ `basename "~{primer_bed}"` == "snap_primers.bed" ]]; then
```
kislyuk, line 550: Consider using quoted subshell calls instead:
```if [[ "$(basename '~{primer_bed}')" == "snap_primers.bed" ]]; then```

kislyuk (approved):
This LGTM overall. It would be good to reword "The SNAP protocol results in primer position offsets due to polymerase...if the data is trimming...SNAP primers, add offset allowance of 5 bp in accordance with ivar developer recommendation" - I'm having trouble parsing the middle part of this sentence, maybe "if the primer bed file given is for the SNAP protocol, then...". It would be best to add an extended explanation of this condition in the comment (similar to the one in the PR description), ideally with a link to an online resource describing this primer slop issue.

In principle, it's not great that we are inferring the protocol in use from the filename of the primers. What if the primer filename changes without the protocol changing? I'm not sure what the best way to deal with that is, but in general I would opt for more explicitly referencing the protocol in the input, instead of inferring it implicitly from an attribute of the input that might change for an unrelated reason.

consensus-genome/run.wdl
katrinakalantar: Thanks!
I've updated the text to be more clear and to include a link to the appropriate upstream PR in iVar that discusses the implications of the change. It's a great point about the issues associated with assuming protocol from the .bed file. I believe we do the same elsewhere, but it would be good to establish a more rigorous scheme or documentation of this.

PR #93, opened 2021-03-31
fix fasta writing bug

morsecodist:
This code was doing custom fasta parsing for reading and writing. I switched the reader to using our fasta iterator here which trims newlines. We had no writer and it was out of scope for the parsing refactor to switch to one so I left the writing as it is. Since there was transformation logic on the read name the newline was added back in so it still produced lines but 1 line per row and malformed. More evidence for why custom parsing can be dangerous. I wanted to limit the scope of this bugfix so I added the newline back instead of doing a fasta overhaul but one is needed.

### DIFF
--- a/short-read-mngs/test/test_short_read_mngs.py
+++ b/short-read-mngs/test/test_short_read_mngs.py
@@ -1,6 +1,15 @@
 import json
 import atexit
+from Bio import SeqIO
 
+def is_valid_fasta(filename):
+    try:
+        with open(filename) as f:
+            for read in SeqIO.parse(f, "fasta"):
+                pass
+        return True
+    except:
+        return False
 
 def test_bench3_viral(short_read_mngs_bench3_viral_outputs):
     # short_read_mngs_bench3_viral_outputs is a fixture defined in ./conftest.py providing the JSON outputs of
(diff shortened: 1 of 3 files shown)


jshoe (approved):
Thanks for fixing! (although I see there is a CI error)

morsecodist:
@jshoe yeah I am not sure why that error is there I can't really understand it tbh. Maybe @mlin or @kislyuk would understand.

kislyuk:
@mlin this CI failure looks like it might be related to https://github.com/chanzuckerberg/idseq-workflows/pull/88 - the test data got moved around?

kislyuk:
Can we add a unit test for this step? Something like https://github.com/chanzuckerberg/idseq-workflows/blob/main/short-read-mngs/test/host_filter/test_RunValidateInput.py?

morsecodist:
@kislyuk yeah I believe so. I don't really get this step but I can do at least a basic one.

short-read-mngs/test/test_short_read_mngs.py
```diff
@@ -20,4 +29,7 @@ def test_bench3_viral(short_read_mngs_bench3_viral_outputs):
     taxa = set(entry["tax_id"] for entry in taxon_counts)
     assert len(taxa) == 149
 
+    for filename in outp["outputs"]:
+        if filename.endswith(".fasta"):
+            assert is_valid_fasta(filename), f"{filename} is not a valid fasta file"
```
morsecodist, line 34: @kislyuk how do you feel about this approach? I realized that step us used in multiple pipeline stages and then I thought maybe file type integrity should just be a global part of our tests. This would have caught that error and it also catches a ton of other parsing errors.

kislyuk: That works great, thanks!

kislyuk (approved):

PR #94, opened 2021-04-06
update to catch vadr errors and associated unit tests

katrinakalantar:
We were seeing sample failure events due to VADR throwing an error for long sample names: `ERROR, at least one sequence name exceeds the maximum GenBank allowed length of 50` 

VADR should not cause the pipeline to fail. Rather, the errors should be caught and propagated in .json while the pipeline continues. This PR addresses these issues as follows:

- Updating the `vadr_options` to use the `--noseqnamemax` flag so that it will not fail for long sample names
- Updating the Vadr step to  include better error handling that will correctly capture the error more accurately and allow the pipeline to continue
-  Added associated unit tests to 1. use the long sample names as a proxy for vadr failures to test the error handling, 2. ensure that the flag successfully eliminates this particular error.

This was tested via unit testing as well as through application to two separate samples.

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -36,7 +36,7 @@ workflow consensus_genome {
         Int normalise  = 1000
         # medaka_model: default is selected to support current ClearLabs workflow
         String medaka_model = "r941_min_high_g360"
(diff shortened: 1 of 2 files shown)


morsecodist (approved):
Thanks for doing this so fast!

morsecodist (approved):

PR #96, opened 2021-04-07
reduce the nanopore workflow min_length parameter to accommodate ClearLabs samples in the interim 

katrinakalantar:
Modify the default `min_length` parameter for nanopore sars-cov-2 consensus genome analysis to achieve higher coverage for ClearLabs samples, which tend to have a distribution of read lengths that include a significant number (> 50%) of reads with lengths below the initial 350 parameter value. 

This PR is to enable DPH users who are actively using the pipeline during the beta phase to achieve genomes of comparable quality. However, further validation and a parameter sweep will need to be done prior to v1 launch to ensure that this modification does not produce issues for standard ONT data.

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -33,7 +33,7 @@ workflow consensus_genome {
         File primer_schemes = "s3://idseq-public-references/consensus-genome/artic-primer-schemes.tar.gz"
         # filters in accordance with recommended parameters in ARTIC SARS-CoV-2 bioinformatics protocol are...
         # ...intended to remove obviously chimeric reads.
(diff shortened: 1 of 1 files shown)


kislyuk (approved):
I'm fine with merging this as is in the name of expediency, but ideally we should have two or more test cases like https://github.com/chanzuckerberg/idseq-workflows/blob/main/consensus-genome/test/test_wdl.py#L120-L130 where the default `min_length` parameter is used followed by an assertion that the average coverage depth reported in output_stats is within a certain range, followed by a custom `min_length` value resulting in a different range.

katrinakalantar:
Given the recent info about ClearLabs pre-filtering, I will close this PR in place of upcoming changes that will skip the `ApplyLengthFilter` altogether for those data.

PR #103, opened 2021-04-15
Call VADR only on SC2 samples

kislyuk:
As per our discussion. This causes the relevant outputs to all be `null` if ``ref_accession_id` is set (a proxy for general CG). Added comments regarding possible future improvements to make VADR run on more viruses and references.


jshoe (approved):
Looks good code-wise!

katrinakalantar (approved):
Yep, this looks good to me. Thank you for documenting the TODO.

PR #82, opened 2021-02-17
Phylo Refactor

morsecodist:

PR #104, opened 2021-04-16
PhyloTree-NG workflow skeleton

kislyuk:
### Work in this PR
- [x] Import @katrinakalantar's and @morsecodist's work in their respective prototypes (https://github.com/katrinakalantar/clustered-phylotree, https://github.com/chanzuckerberg/idseq-workflows/pull/82)
- [x] Add a placeholder test suite and documentation (see `FIXME` marker)
### Remaining work
- [x] Confirm I/O signature: do we need `Superkingdom Name` and do we need to adjust the k-mer size according to it as we did for ksnp3?
- [ ] Confirm I/O signature: is the primary `reference` and secondary `additional_references` logic correct? If so, do we need to support taxon IDs, accession IDs, or both in either case?
- [ ] Add a meaningful test sample
- [ ] Add tests
- [ ] Check and validate logic and error handling
- [ ] Scientific validation


phylotree-ng/analysis_support/simulate_low_coverage_genome.py
```diff
@@ -0,0 +1,56 @@
...
+    while i < ns_to_create_sim_cov:
+        position += 1
+        if new_sequence[start_pos + position] == 'N':
+            continue
+        else:
+            new_sequence[start_pos + position] = 'N'
+            i += 1
+
+    final_coverage = 1 - Counter(new_sequence)['N']/len_of_seq
+    print("final coverage: " + str(final_coverage))
```
morsecodist, line 51: nitpick: `print("final coverage:", final_coverage)`

phylotree-ng/analysis_support/simulate_low_coverage_genome.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+for simulated_coverage in [.90, .75, .5, .25]:
+    print(simulated_coverage)
+    delta_to_simulated_coverage = existing_coverage - simulated_coverage
+    ns_to_create_sim_cov = round(delta_to_simulated_coverage*len_of_seq)
+    print(ns_to_create_sim_cov)
+    start_pos = randint(0, (len_of_seq-ns_to_create_sim_cov-nt_counts['N']))
+
+    print('desired simulated cov: ' + str(simulated_coverage))
```
morsecodist, line 35: nitpick: `print('desired simulated cov:', simulated_coverage)`

phylotree-ng/analysis_support/simulate_low_coverage_genome.py
```diff
@@ -0,0 +1,56 @@
...
+    print('existing cov: ' + str(existing_coverage))
```
morsecodist, line 36: nitpick: `print('existing cov:', existing_coverage)`

phylotree-ng/run.wdl
```diff
@@ -0,0 +1,427 @@
...
+    # associated with them. Emit one fasta file per reference, to be graphed in a phylotree along with samples
+    input {
+        Array[ReferenceInfo] references
+        String docker_image_id
+    }
+
+    command <<<
+    for accession_id in $(jq -r .[].accession_id "~{write_json(references)}"); do
+        taxoniq get-from-s3 --accession-id $accession_id > $accession_id.fasta
+    done
```
morsecodist, line 122: This is awesome!

phylotree-ng/run.wdl
```diff
@@ -0,0 +1,427 @@
...
+
+    # cluster the data, create a dendrogram
+    Z = cluster.hierarchy.linkage(1-df3, method='complete')
+    dn = cluster.hierarchy.dendrogram(Z, leaf_rotation=90, labels=df3.index)
+    plt.savefig('dendrogram.png', bbox_inches='tight')
+
+    trim_height = "~{cut_height}"
+    cutree = cluster.hierarchy.cut_tree(Z, height=float(trim_height))
+    ordered_clusterids = [i[0] for i in cutree]
+    cluster_assignments = dict(zip(df3.index,ordered_clusterids ))
```
morsecodist, line 226: nit (rogue space): `zip(df3.index, ordered_clusterids)`

phylotree-ng/run.wdl
```diff
@@ -0,0 +1,427 @@
...
+    stats = {"sample_name": "insert_sample_name"}
+
+    # write cluster contents to files for future processing
+    for c in cluster_sets.keys():
+        stats[c] = ' '.join(cluster_sets[c]) # record cluster IDs in stats file for all clusters
+        if(len(cluster_sets[c])) > 2: # only output files where there are > 2 samples
+            filenames = '\n'.join(cluster_sets[c])
+            text_file = open("./cluster_files/cluster_" + str(c), "w")
+            n = text_file.write(filenames)
+            text_file.close()
```
morsecodist, line 242:
```python
with open("./cluster_files/cluster_" + str(c), "w") as text_file:
    n = text_file.write(filenames)
```

phylotree-ng/run.wdl
```diff
@@ -0,0 +1,427 @@
...
+
+    #import seaborn as sns
```
morsecodist, line 244: delete line

phylotree-ng/run.wdl
```diff
@@ -0,0 +1,427 @@
...
+        split_line = l.split()
+        if line_id == 1:    # on the first line, add entry to both columns to ensure square distance matrix
+            new_line = [split_line[0], split_line[0], 0, 0, 0, 0, 0, 0]
+            processed_lines.append(new_line)
+        processed_lines.append(split_line)
+        line_id += 1
+    processed_lines.append([split_line[1], split_line[1], 0, 0, 0, 0, 0, 0])  # on the last line, add entry to both columns
+
+    # create dataframe from the distance data
+    df = pd.DataFrame.from_dict(dict(zip([i for i in range(len(processed_lines))], processed_lines)), orient='index')
```
morsecodist, line 202: This confused me a bit I think it may be clearer and possibly a slightly smaller memory footprint to replace this:

```python
# open the ska distance file
distances_file = open("~{ska_distances}", 'r')
lines = distances_file.readlines()

# process the lines (cannot just import with pandas due to non-standard whitespace issues)
processed_lines = []
line_id = 1
for l in lines[1:]:
    split_line = l.split()
    if line_id == 1:    # on the first line, add entry to both columns to ensure square distance matrix
        new_line = [split_line[0], split_line[0], 0, 0, 0, 0, 0, 0]
        processed_lines.append(new_line)
    processed_lines.append(split_line)
    line_id += 1
processed_lines.append([split_line[1], split_line[1], 0, 0, 0, 0, 0, 0])  # on the last line, add entry to both columns

# create dataframe from the distance data
df = pd.DataFrame.from_dict(dict(zip([i for i in range(len(processed_lines))], processed_lines)), orient='index')
```

With this:

```python
def processed_lines():
    with open("~{ska_distances}", 'r') as distances:
        for i, line in enumerate(distances):
            split_line = line.split()
            if i == 1:
                # on the first line, add entry to both columns to ensure square distance matrix
                yield [split_line[0], split_line[0], 0, 0, 0, 0, 0, 0]
            elif i > 1: # 0 is purposefully omitted
                yield split_line
       # on the last line, add entry to both columns
        yield [split_line[1], split_line[1], 0, 0, 0, 0, 0, 0]

# create dataframe from the distance data
df = pd.DataFrame.from_dict(dict(enumerate(processed_lines())), orient='index')
```

morsecodist (approved):
Looks great overall. A ton of progress here! Just a few minor python comments.

phylotree-ng/Phylotree.md
```diff
@@ -0,0 +1,117 @@
+# Phylo Tree Description
```
katrinakalantar, line 1: Thanks for adding Todd's documentation! 
It may be nice to add a note that this is a work-in-progress and the code / docs will come into alignment over the development cycle but currently are out of sync.

phylotree-ng/analysis_support/simulate_low_coverage_genome.py
```diff
@@ -0,0 +1,56 @@
+'''
```
katrinakalantar, line 1: Should we keep the `analysis_support` scripts in this repo? I'm not sure if they'll be useful in production. Their only utility thus far has been for generating test datasets.

phylotree-ng/test/test_wdl.py
```diff
@@ -0,0 +1,22 @@
...
+# import json
+
+from test_util import WDLTestCase
+
+
+class TestPhylotree(WDLTestCase):
+    wdl = os.path.join(os.path.dirname(__file__), "..", "run.wdl")
+    common_inputs = {
+        "samples": [{"sample_name": "test_sample", "workflow_run_id": 12345}],
+        "reference": {"taxon_id": 482},
```
katrinakalantar, line 13: Question just for my own understanding - will these end up being common inputs? Or are they just to show a skeleton of how common inputs will be added?

phylotree-ng/run.wdl
```diff
@@ -0,0 +1,425 @@
...
+    Int? taxon_id
+    String? accession_id
+}
+
+workflow phylotree {
+    input {
+	Array[SampleInfo] samples
+        ReferenceInfo reference
+
+        # String superkingdom_name - TODO: is this needed?
```
katrinakalantar, line 20: I hadn't incorporated different kmer length parameters for different superkingdoms yet, but I think we should inherit the defaults from the kSNP3 pipeline to start and can do more work to tune those later. So it would be good to include superkingdom_name as an input.

phylotree-ng/run.wdl
katrinakalantar: This makes sense to me, thanks for the suggestion on how to improve this! 
One question on the suggested code - would we then want to say `i == 0` and ` i > 0`? I'm assuming the `enumerate(distances)` is zero-based in which case starting to emit data at `i > 1` might make us lose a line. 

While I do think v0 will not include the full clustering functionality (as noted above), it would be great to still parse the distances file from long format into a wide format distance matrix (what this code chunk is doing) so that we could return a heatmap output.

katrinakalantar (approved):
This looks like a solid structure to me! Thanks for putting together this skeleton. Additionally, thanks @morsecodist for some of the python suggestions. 

I added a few notes and a few questions, but nothing blocking.

phylotree-ng/run.wdl
morsecodist: I think the existing code skips the first line (`lines[1:]`) and converts the second line (index 1) to the zeroed version, does the rest of the lines normally, then adds the last line a second time with zeros. Was that your intention? Maybe we can quickly sync on your intent with this block.

kislyuk: @morsecodist @katrinakalantar I replaced the code with Todd's suggestion and filed https://app.clubhouse.io/idseq/story/133579/follow-up-on-computeclusters-logic

phylotree-ng/Phylotree.md
kislyuk: Added a comment:

> # PhyloTree Workflow: Technical Description

> The following documentation applies to the legacy kSNP3 PhyloTree workflow. The new SKA-based PhyloTree-NG workflow is a work in progress. The documentation below will be adapted to the new workflow over time.

phylotree-ng/analysis_support/simulate_low_coverage_genome.py
kislyuk: I think it's useful to keep auxiliary code like this in version control next to the tests. This repo is responsible not just for production artifacts but also for supporting our workflows with tests and documentation.

phylotree-ng/test/test_wdl.py
kislyuk: Here "common" refers to inputs that are common across all the tests (not all the production invocations of the workflow), and in this particular case it's just meant as an example to get the I/O plumbing demonstrated - it's not meant to be scientifically meaningful, instead it's meant to be replaced with meaningful inputs by the implementer.

phylotree-ng/run.wdl
kislyuk: Great! Added `superkingdom_name` to the workflow signature, and a TODO

phylotree-ng/run.wdl
```diff
@@ -0,0 +1,425 @@
...
+        docker_image_id = docker_image_id
+    }
+
+    call RunSKA {
+        input:
+        sample_and_reference_fastas = flatten([GetSampleContigFastas.sample_contig_fastas, GetReferenceFastas.reference_fastas]),
+        docker_image_id = docker_image_id
+    }
+
+    call ComputeClusters {
```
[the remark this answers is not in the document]
kislyuk: OK - I will leave it to the implementer to comment out the calls and emitted output references for ComputeClusters.

PR #106, opened 2021-04-20
Bump taxoniq to v0.6.0

kislyuk:
Resolves a bug where taxoniq would report some sequences to be 1, 2, or 3 nt shorter than they really are (due to the intricacies of the NCBI BLAST nucleotide DB format).


katrinakalantar (approved):
This looks good to me.

kislyuk:
Resolves a bug where taxoniq would report some sequences to be 1, 2, or 3 nt shorter than they really are (due to the intricacies of the NCBI BLAST nucleotide DB format).

PR #107, opened 2021-04-23
fix reading ska tsv

morsecodist:
After syncing with @katrinakalantar I realize I misinterpreted this code and introduced a logical error. I also realized all this code could be replaced with panda's built-in methods. I think this makes things much simpler and more performant. I tested locally on a file @katrinakalantar provided and it worked. She mentioned some tsv parsing weirdness with these files so we can keep an eye out when testing the workflow as a whole and address any issues as they arise.

### DIFF
--- a/phylotree-ng/run.wdl
+++ b/phylotree-ng/run.wdl
@@ -183,6 +183,8 @@ task ComputeClusters {
     import seaborn as sns
     import math
 
+    # we have observed some strange parsing behavior of this file, ensure it works with end to end testing
+    # we may need to use a regex separator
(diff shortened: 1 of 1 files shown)


katrinakalantar (approved):
Looks good to me!  Thanks for simplifying the logic with pandas and noting that we'll keep an eye out for weirdness with SKA .tsv output files in end-to-end testing.

morsecodist:
@katrinakalantar thanks actually I will add a comment in the code as well. If there is weirdness we can use pandas' regex separator so the code won't change too much.

PR #110, opened 2021-05-04
Add MN908947.3.fa test file

jshoe:
- Copied this from `s3://idseq-public-references/consensus-genome/MN908947.3.fa`, intending to use this in a system test.


kislyuk:
Other test fixtures in the system test (input fastqs) are passed by reference as github raw URLs in the idseq-workflows repo.

Consider either making the approach in this PR consistent with that approach (pass the input URL by reference instead of staging it in the source code), or copying the other test fixtures into this repo as well.

jshoe:
@kislyuk Not sure I completely understand the comment: I don't think we have this file (`MN908947.3.fa`) in the repos yet, so I was adding it here in idseq-workflows to follow that pattern. Unless we wanted to reference `s3://idseq-public-references/consensus-genome/MN908947.3.fa` directly ?

kislyuk:
Sorry about the confusion, I meant referencing the file directly, yes.

PR #115, opened 2021-05-17
emit star logfile

morsecodist:
Move one of star's outputs into the output directory and add it as an output in the WDL.

Includes a test for file presence to verify the file is being transferred.

I want to refactor the star step to dramatically simplify it but that is a larger task than I have time for this sprint so I opted to change as little as possible without making things worse.


jshoe (approved):
LGTM

morsecodist:
Just a small test issue, fixed and merging

PR #117, opened 2021-05-18
remove nt.refseq output

morsecodist:
I was going to fix this by combining steps but recently I have done some experimenting and it seems the viz is robust to changes in outputs. I think it is simpler and cleaner to just remove this as an output to save space. This is also a bit of a trial run for the valid inputs removal which needs to be done this way.


jshoe (approved):

morsecodist:
The failure on this PR is a spurious consensus genome failure, rerunning

PR #118, opened 2021-05-18
optimize github actions jobs to avoid redundant runs

morsecodist:
Our workflows are SUPER long now so I added some stuff to prevent them from running in certain contexts:

- New workflows cancel previously running workflows to keep stuff from getting clogged
- CI only runs on the directories you have made changes in
- If linting fails the other CI jobs are cancelled
- Only run the mngs benchmark if there is mngs diff

I experimented with the different diffs to show it is working. I added and removed some mngs diff which is why the benchmarks ran on the last commit.

### DIFF
--- a/.github/workflows/short-read-mngs-viral-benchmarks.yml
+++ b/.github/workflows/short-read-mngs-viral-benchmarks.yml
@@ -16,8 +16,19 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
+  has-diff:
+    runs-on: ubuntu-20.04
+    outputs:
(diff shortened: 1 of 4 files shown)


tfrcarvalho:
Nit: can you make the title of the PR more specific? You scared me a bit there...

Cancelation is per branch, right?

morsecodist:
@tfrcarvalho sorry done, and yes I confirmed it is per branch

tfrcarvalho:
Looks great, thanks!

tfrcarvalho (approved):

PR #119, opened 2021-05-19
quote clusters csv

morsecodist:
Instead of filtering out characters (I notice some example fasta/fastqs do use special characters) I am surrounding every cell with `'`. I modify our parser function to use this as a quotecharacter so this should be safer. Added a test as well.

### DIFF
--- a/.github/workflows/wdl-ci.yml
+++ b/.github/workflows/wdl-ci.yml
@@ -49,7 +49,7 @@ jobs:
           fetch-depth: 2
       - id: list_dirs
         # only run on wdl dirs with diff
-        run: echo "::set-output name=matrix::$(ls */*.wdl|xargs dirname|xargs git diff --name-status HEAD^|cut -c 3-|xargs dirname|sort|uniq|jq -cnR '[inputs|select(length>0)]')"
+        run: echo "::set-output name=matrix::$(ls */*.wdl|xargs dirname|xargs git diff --name-status HEAD^|cut -c 3-|xargs dirname|sed 's/\/.*$//g'|sort|uniq|jq -cnR '[inputs|select(length>0)]')"
 
   wdl-ci:
     runs-on: ubuntu-20.04
--- a/.github/workflows/wdl-ci.yml
+++ b/.github/workflows/wdl-ci.yml
@@ -45,6 +45,8 @@ jobs:
       matrix: ${{steps.list_dirs.outputs.matrix}}
     steps:
       - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/wdl-ci.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           fetch-depth: 2
       - id: list_dirs
         # only run on wdl dirs with diff
-        run: echo "::set-output name=matrix::$(ls */*.wdl|xargs dirname|xargs git diff --name-status HEAD^|cut -c 3-|xargs dirname|sort|uniq|jq -cnR '[inputs|select(length>0)]')"
+        run: echo "::set-output name=matrix::$(ls */*.wdl|xargs dirname|xargs git diff --name-status HEAD^|cut -c 3-|xargs dirname|sed 's/\/.*$//g'|sort|uniq|jq -cnR '[inputs|select(length>0)]')"
```
morsecodist, line 52: Fix issue with my recent change to the github actions

short-read-mngs/idseq-dag/idseq_dag/steps/run_idseq_dedup.py
```diff
@@ -60,6 +63,12 @@ def run(self):
             )
         )
 
+        # quote the csv with ' to guard against potential csv injection
+        with open(duplicate_clusters_raw_path) as r, open(duplicate_clusters_path, 'w') as w:
```
tfrcarvalho, line 67: No integers here, right? Just a map of reads to a cluster read?

morsecodist: Yeah it is just a map of ID to ID, both strings

tfrcarvalho (approved):

short-read-mngs/idseq-dag/idseq_dag/steps/run_idseq_dedup.py
```diff
@@ -60,6 +63,12 @@ def run(self):
...
+            writer = csv.writer(w)
+            for row in csv.reader(r):
+                writer.writerow(f"'{elem}'" for elem in row)
```
jshoe, line 70: I think just prepending the single quote is a little better because the users will see the trailing quote:

Screen Shot 2021-05-19 at 9 48 15 AM

https://owasp.org/www-community/attacks/CSV_Injection "Alternatively, prepend each cell field with a single quote, so that their content will be read as text by the spreadsheet editor"

PR #120, opened 2021-05-19
switch to initial space safety

morsecodist:
@jshoe pointed out that my implementation degraded user experience. This should be a better solution that balances user experience with parse-ability.

New tests too.

BTW the pipeline integration tests assure that the IDs are being found by steps since if one is ever not found the pipeline will crash.


jshoe (approved):
LGTM

Issue #97, opened 2021-04-08
possible issue with INDEL calling by samtools


danrlu:
We recently discovered a [bug](https://github.com/czbiohub/sc2-illumina-pipeline/issues/80) in our code that by default we were leaving out most of the INDELs in vcf.

I see idseq is not enforcing the samtools version, so I'm not sure whether it would be the case for the version that is used [here](https://github.com/chanzuckerberg/idseq-workflows/blob/2a08bf11210cc85db00af87f24f9e0235d331ec2/consensus-genome/run.wdl#L686). FYR

kislyuk:
Hi @danrlu, thanks for letting us know. IDseq does indeed set a specific samtools version, by installing the version that ships with Ubuntu 20.04 LTS in the Dockerfile for the workflow (https://github.com/chanzuckerberg/idseq-workflows/blob/main/consensus-genome/Dockerfile). The specific version can be found at https://packages.ubuntu.com/focal/samtools - it's version 1.10.

It's not clear to me from reading https://github.com/czbiohub/sc2-illumina-pipeline/issues/80 and https://github.com/czbiohub/sc2-illumina-pipeline/pull/81 which samtools versions have the issue you refer to, and whether we need to add a custom `-L` setting when calling `samtools mpileup`. Can you clarify? cc @katrinakalantar

danrlu:
Oh I didn't know the version can be specified this way, very good to learn! Sorry to leave out the critical piece of info... our samtools is [1.9](https://github.com/czbiohub/sc2-illumina-pipeline/blob/d7f3a8a0ef34fe53e9008d538cb324a2b3920197/environment.yaml#L9). `mpileup` is a samtools function that is now moved into bcftools, and I tested samtools 1.9 and bcftools 1.11, both of which need the `-L` flag. I passed all related info and files to Katrina.

PR #122, opened 2021-05-25
Update samtools parameters to improve indel calling

katrinakalantar:
A recent issue was raised that we were missing some INDEL calls in the output .vcf files https://github.com/chanzuckerberg/idseq-workflows/issues/97

This PR does the following:

- addresses the issue by modifying the `CallVariants` step to use the `-L` and `-d` parameters to improve sensitivity for INDELS in the output .vcf files.
- passes the quality threshold that was used for `CallConsensus` to also be used in `CallVariants` consistent with recent changes upstream in the CZB sc2 pipeline. 

This was tested offline using a sample with a known INDEL. The pipeline previously did not output the corresponding lines in the .vcf file, but the change has resulted in successful identification of the known INDEL.

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -49,7 +49,7 @@ workflow consensus_genome {
         Boolean trim_adapters = true
 
         Float ivarFreqThreshold = 0.9
-        Int   ivarQualTreshold  = 20
+        Int   ivarQualThreshold  = 20
         Int   minDepth          = 10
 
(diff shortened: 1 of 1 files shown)


morsecodist:
Did you want me to review this?

morsecodist (changes requested):
You missed a usage of `ivarQualTreshold` on line 152. This needs to be replaced with `ivarQualThreshold`

morsecodist (approved):
Looks good!

PR #124, opened 2021-05-26
remove nr-refseq output

morsecodist:
I removed nt.refseq already. This file isn't as big but for consistency I think we should remove it too.


katrinakalantar (approved):
This looks good to me!

PR #127, opened 2021-06-03
short-read-mngs RunAssembly: contig length filter

mlin:
Apply a 100-nucleotide length filter to the contigs as they emerge from SPAdes, to reduce the burden on subsequent steps like BlastContigs. Shorter contigs can make up a majority of SPAdes output currently (sample-dependent), yet seem of minimal value considering they're shorter than the input reads.

The downfiltered contigs are used for all subsequent steps (including the Bowtie2 realignment of reads to contigs immediately following SPAdes). The original set is copied into a new workflow output, `assembly_out_assembly_contigs_all_fasta`, which remains unused.

### DIFF
--- a/short-read-mngs/idseq-dag/idseq_dag/steps/run_assembly.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/steps/run_assembly.py
@@ -6,6 +6,7 @@ from collections import defaultdict
 from idseq_dag.engine.pipeline_step import PipelineStep
 import idseq_dag.util.command as command
 import idseq_dag.util.command_patterns as command_patterns
+import idseq_dag.util.fasta as fasta
 
 from idseq_dag.util.m8 import MIN_CONTIG_SIZE
 from idseq_dag.util.count import get_read_cluster_size, load_duplicate_cluster_sizes, READ_COUNTING_MODE, ReadCountingMode
@@ -64,11 +65,13 @@ class PipelineStepRunAssembly(PipelineStep):
         duplicate_cluster_sizes_path, = self.input_files_local[1]
         assert duplicate_cluster_sizes_path.endswith(".tsv"), self.input_files_local[1]
 
-        assembled_contig, assembled_scaffold, bowtie_sam, contig_stats = self.output_files_local()
+        assembled_contig, assembled_contig_all, assembled_scaffold, bowtie_sam, contig_stats = self.output_files_local()
         read2contig = {}
         memory = self.additional_attributes.get('memory', 100)
-        self.assemble(input_fasta, input_fasta2, bowtie_fasta, duplicate_cluster_sizes_path, assembled_contig, assembled_scaffold,
-                      bowtie_sam, contig_stats, read2contig, int(memory))
+        min_contig_length = int(self.additional_attributes.get('min_contig_length', 0))
+        self.assemble(input_fasta, input_fasta2, bowtie_fasta, duplicate_cluster_sizes_path,
+                      assembled_contig, assembled_contig_all, assembled_scaffold,
+                      bowtie_sam, contig_stats, read2contig, int(memory), min_contig_length)
 
     @staticmethod
     def assemble(input_fasta,
@@ -76,11 +79,13 @@ class PipelineStepRunAssembly(PipelineStep):
                  bowtie_fasta,  # fasta file for running bowtie against contigs
                  duplicate_cluster_sizes_path,
                  assembled_contig,
+                 assembled_contig_all,
                  assembled_scaffold,
                  bowtie_sam,
                  contig_stats,
                  read2contig,
-                 memory=100):
+                 memory=100,
+                 min_contig_length=0):
         basedir = os.path.dirname(assembled_contig)
         assembled_dir = os.path.join(basedir, 'spades')
         command.make_dirs(assembled_dir)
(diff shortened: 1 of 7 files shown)


short-read-mngs/postprocess.wdl
```diff
@@ -659,6 +661,7 @@ workflow idseq_postprocess {
 
   output {
     File assembly_out_assembly_contigs_fasta = RunAssembly.assembly_contigs_fasta
+    File assembly_out_assembly_contigs_all_fasta = RunAssembly.assembly_contigs_all_fasta
```
katrinakalantar, line 666: Great that we're still outputting the full assembled contig file!

short-read-mngs/test/README.md
```diff
@@ -26,5 +26,5 @@ Then, build the docker image from the current code revision and start the tests:
 ```bash
 cd idseq-workflows
 docker build -t idseq-short-read-mngs short-read-mngs
-make test-short-read-mngs
+DOCKER_IMAGE_ID=idseq-short-read-mngs make test-short-read-mngs
```
katrinakalantar, line 29: Should these be a single line?

short-read-mngs/idseq-dag/idseq_dag/util/fasta.py
```diff
@@ -13,19 +13,23 @@ class Read(NamedTuple):
...
+            if not line:
                 break
-            # the performance penalty for these asserts is only 8 percent
-            assert header[0] == '>'
-            assert sequence[0] != '>'
-            # the performance penalty for constructing a Read tuple is 40 percent
-            yield Read(header, sequence)
+            if line[0] == '>':
+                if header:
+                    yield Read(header, ("".join(sequence_wip) if len(sequence_wip) != 1 else sequence_wip[0]))
```
katrinakalantar, line 25: Is there a particular reason to include the `if len(sequence_wip) != 1` statement? I think the same result would be achieved if `"".join(sequence_wip)` was used even for cases where `sequence_wip` only has length = 1, but am not certain of if this was done for efficiency purposes.

short-read-mngs/test/postprocess/test_RunAssembly.py
```diff
@@ -0,0 +1,68 @@
...
+    # count fasta
+    assembly_contigs_fasta = os.path.join(
+        outp["dir"], outp["outputs"][f"{task_name}.assembly_contigs_fasta"]
+    )
+    assembly_contigs_all_fasta = os.path.join(
+        outp["dir"], outp["outputs"][f"{task_name}.assembly_contigs_all_fasta"]
+    )
+    contigs_headers = fasta_headers(assembly_contigs_fasta)
+    contigs_all_headers = fasta_headers(assembly_contigs_all_fasta)
+    assert contigs_headers and (set(contigs_all_headers) - set(contigs_headers))
```
katrinakalantar, line 57: optional: for completeness, should we additionally assert on the length of a sequence in `set(contigs_all_headers) - set(contigs_headers)` to ensure length is < `min_contig_length`?

katrinakalantar (commented):
This looks good to me, I added a few minor comments. 
I'll let @morsecodist give the final approval.

short-read-mngs/idseq-dag/idseq_dag/steps/run_assembly.py
```diff
@@ -124,14 +129,25 @@ def assemble(input_fasta,
...
                     )
                 )
-            command.move_file(assembled_contig_tmp, assembled_contig)
+            command.move_file(assembled_contig_tmp, assembled_contig_all)
             command.move_file(assembled_scaffold_tmp, assembled_scaffold)
 
+            if min_contig_length:
+                # apply contig length filter
+                with open(assembled_contig, "w") as outfile:
+                    for record in fasta.iterator(assembled_contig_all):
```
morsecodist, line 138: I am trying to move us away from custom parsing. We already have bio python as a dependency. Could you replace with something like:

```python
from Bio import SeqIO

SeqIO.write(
    (r for r in SeqIO.parse(assembled_contig_all, "fasta") if len(r.seq) >= min_contig_length),
    assembled_contig,
    "fasta",
)
```

This would make your changes in short-read-mngs/idseq-dag/idseq_dag/util/fasta.py unnecessary

short-read-mngs/idseq-dag/idseq_dag/util/fasta.py
```diff
@@ -13,19 +13,23 @@ class Read(NamedTuple):
 
 def iterator(fasta_file: str) -> Iterator[Read]:
```
morsecodist, line 14: As mentioned in my other comment these changes might be unnecessary

short-read-mngs/postprocess.wdl
```diff
@@ -6,6 +6,7 @@ task RunAssembly {
     String s3_wd_uri
     Array[File] host_filter_out_gsnap_filter_fa
     File duplicate_cluster_sizes_tsv
+    Int min_contig_length
```
morsecodist, line 9: Can you set the default here and make it optional so it is more obvious and we could disable it more cleanly if we ever support that?

```
Int? min_contig_length = 100
```

Assuming we can explicitly pass in null here.

short-read-mngs/idseq-dag/idseq_dag/steps/run_assembly.py
```diff
@@ -64,23 +65,27 @@ def run(self):
         duplicate_cluster_sizes_path, = self.input_files_local[1]
         assert duplicate_cluster_sizes_path.endswith(".tsv"), self.input_files_local[1]
 
-        assembled_contig, assembled_scaffold, bowtie_sam, contig_stats = self.output_files_local()
+        assembled_contig, assembled_contig_all, assembled_scaffold, bowtie_sam, contig_stats = self.output_files_local()
         read2contig = {}
         memory = self.additional_attributes.get('memory', 100)
-        self.assemble(input_fasta, input_fasta2, bowtie_fasta, duplicate_cluster_sizes_path, assembled_contig, assembled_scaffold,
-                      bowtie_sam, contig_stats, read2contig, int(memory))
+        min_contig_length = int(self.additional_attributes.get('min_contig_length', 0))
```
morsecodist, line 71: See comment about moving the default. Also here it looks like the default is no filtering rather than 100?

morsecodist (changes requested):
Looks great overall. Love the tests. Just a few small tweaks.

short-read-mngs/postprocess.wdl
mlin: I'm setting the default at the workflow-level input rn: https://github.com/chanzuckerberg/idseq-workflows/pull/127/files#diff-d2e6572472ef03a6e96e659a5de68aaf3c3954db41875231be060bbaf4e3a8c0R484

There's actually an ambiguity in the WDL spec we're [working to clarify right now](https://github.com/openwdl/wdl/pull/464) that makes it awkward to set the default value inside the task, but also expose a workflow-level input to optionally override it. It's funny it comes up organically here, as we were just having a healthy debate about that PR last week.

If we needed to, we'd set the workflow-level input to 0.

short-read-mngs/test/README.md
mlin: That just makes the definition effective only for this invocation instead of the rest of the shell session (since `DOCKER_IMAGE_ID` is a pretty generic name)

short-read-mngs/idseq-dag/idseq_dag/steps/run_assembly.py
mlin: Agreed and implemented. I found I had to add biopython to the idseq-dag setup.py requires, I guess the production Dockerfile installs it separately.

short-read-mngs/idseq-dag/idseq_dag/util/fasta.py
mlin: (obviated)

morsecodist (approved):
Looks good thanks!

PR #129, opened 2021-06-04
Upgrade vadr model to vadr-models-sarscov2-1.2-2.tar.gz

morsecodist:
Tests may fail due to space issue.


morsecodist:
The tests are failing due to the space issue. I will fix that but I don't want it to be a blocker for this change. I am running them locally and when they pass I will call it out here and we can merge.

rzlim08 (approved):
Looks good!

PR #130, opened 2021-06-09
Test various medaka models

rzlim08:


consensus-genome/test/test_wdl.py
```diff
@@ -146,6 +146,33 @@ def test_sars_cov2_ont_cg(self):
...
+        Test that the pipeline will fail if the medaka model is incompatible
+        """
+        model = "r941_prom_snp_g360"
+        fastq = os.path.join(os.path.dirname(__file__), "no_host_1.fq.gz")
+        args = ["prefix=''", "sample=test_sample", f"fastqs={fastq}",
+                "normalise=1000", f"medaka_model={model}",
+                "primer_schemes=s3://idseq-public-references/consensus-genome/artic-primer-schemes.tar.gz"]
+        with self.assertRaises(CalledProcessError) as ecm:
+            self.run_miniwdl(args, task="RunMinion")
+            self.assertRunFailed(ecm, task="RunMinion", error="CalledProcessError")
```
morsecodist, line 174: I wonder if this line ever actually gets reached. Wouldn't this throw an error on line 173 then the assertion on 172 passes and the test passes? Sorry it is a pain to re-run (maybe you can rebase with my test reliability change) but could you try running without this?

consensus-genome/test/test_wdl.py
```diff
@@ -146,6 +146,33 @@ def test_sars_cov2_ont_cg(self):
...
+        Test that the pipeline will run a variety of different medaka models
+        """
+        models = ["r10_min_high_g340", "r103_min_high_g345", "r941_prom_fast_g303"]
+        fastq = os.path.join(os.path.dirname(__file__), "no_host_1.fq.gz")
+        for model in models:
+            args = ["prefix=''", "sample=test_sample", f"fastqs={fastq}",
+                    "normalise=1000", f"medaka_model={model}",
+                    "primer_schemes=s3://idseq-public-references/consensus-genome/artic-primer-schemes.tar.gz"]
+            res = self.run_miniwdl(args, task="RunMinion")
+            for _, filename in res["outputs"].items():
```
morsecodist, line 160: nit:

```python
for _, filename in res["outputs"].items():
```

```python
for filename in res["outputs"].values():
```

morsecodist (changes requested):
Minor comments. Also 1.03MB is pretty big to commit. Would it be possible to get the same results by subsampling your input file?

consensus-genome/test/test_wdl.py
rzlim08: You're totally right, I over indented here. As a result, I found out that the error returned is not well handled by miniwdl so self.assertRunFailed can't parse the outputs. So I replaced the call with some of the internal logic that should work.

consensus-genome/test/test_wdl.py
rzlim08: Fixed!

rzlim08:
> Minor comments. Also 1.03MB is pretty big to commit. Would it be possible to get the same results by subsampling your input file?

Done! Got it down to 129KB, could probably get down lower if needed.

morsecodist (approved):
Looks good thanks for the fixes!

PR #131, opened 2021-06-11
fix bug with NA values in ska.distances.tsv and add known user error for divergent samples

katrinakalantar:
There are cases where NA values will appear in the ska.distances.tsv file for samples with no overlapping kmers. We were filling `na` values with zeroes at a later step in the `ComputeClusters` script, but this gives the false appearance that these samples are extremely similar. This update applies `fillna = 1` earlier in the parsing of the `ska.distances.tsv` file so that the samples with NA values are maximally divergent in the resulting heatmap.

Additionally, cases where NA values appear in the `ska.distances.tsv` file may result in the case where there are no variants from which to construct the tree. This causes an error due to invalid inputs to iqtree (i.e. a malformed .fasta file). For example:
```
>seq1

>seq2

>seq3

```

This PR also adds functionality to catch such errors under the class of `TooDivergentError`.

This was tested on a sample set that generated NA values in the `ska.distances.tsv` file.

### DIFF
--- a/phylotree-ng/python_steps/compute_clusters.py
+++ b/phylotree-ng/python_steps/compute_clusters.py
@@ -26,6 +26,7 @@ def main(ska_distances: str, trim_height: float, samples: Iterable[Sample], outp
     # we may need to use a regex separator
     df = pd.read_csv(ska_distances, sep='\t')
     df.reset_index(drop=True, inplace=True)
+    df.fillna(1, inplace=True) # fill NA values for SNP and Mash-like Distance with 1 (maximum dist)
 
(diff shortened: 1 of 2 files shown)


morsecodist (approved):
Do you think we need some sort of fasta filetype validation?

katrinakalantar:
@morsecodist fasta filetype validation could improve our ability to transmit debuggable errors to the dev team (wouldn't make a difference from a user-interpretation side), but if you think it would help with maintainability then we should definitely do that. Do you think that should be wrapped into this PR?

PR #133, opened 2021-06-14
Add option for primer set 

rzlim08:
I had to add a couple of input files to step test Quast, which are in "consensus-genome/test/quast", I can reorganize though if preferred.


morsecodist (approved):
`consensus-genome/test/no_host_1.fq.gz` is empty is that intentional?

Looks good overall

rzlim08:
@morsecodist 
> `consensus-genome/test/no_host_1.fq.gz` is empty is that intentional?
I think the commit might be empty, since the file is already in main, but I think the file is still there. I think this showed up because of a weird rebase that I did.

PR #138, opened 2021-06-17
add new RealignConsensus step

katrinakalantar:
This PR addresses the pipeline changes needed to meet a user request to re-align the consensus genome to the reference to aid in identification of INDELS. The realignment step is already performed in the RunMinion step for ONT data, but was not performed for Illumina data. This PR exposes the existing results for ONT and adds a new step to generate the results for Illumina data.

The results will not be available to users with this PR - that would require additional work on the web side to make the results available.

### DIFF
--- a/consensus-genome/Dockerfile
+++ b/consensus-genome/Dockerfile
@@ -28,6 +28,7 @@ RUN apt-get -qq update && apt-get -qq -y install \
   freebayes \
   kraken2 \
   minimap2 \
+  muscle \
   samtools \
   seqtk \
   trim-galore \
--- a/consensus-genome/Dockerfile
+++ b/consensus-genome/Dockerfile
@@ -46,6 +46,7 @@ RUN apt-get -qq update && apt-get -qq -y install \
   automake \
   tabix \
   fasta3 \
+  wget \
   && locale-gen en_US.UTF-8
 
 # These newer versions of infernal and ncbi-blast+ are required by VADR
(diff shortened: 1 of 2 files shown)


consensus-genome/Dockerfile
```diff
@@ -98,6 +99,9 @@ RUN /bin/bash -c "set -e; mkdir -p /usr/local/lib/site_perl; \
   VADREASELDIR=/usr/local/bin \
   VADRBIOEASELDIR=/usr/local/bin > /etc/profile.d/vadr.sh"
 
+RUN wget https://www.drive5.com/muscle/downloads3.8.31/muscle3.8.31_i86linux32.tar.gz \
+&& tar zxvf muscle3.8.31_i86linux32.tar.gz && mv muscle3.8.31_i86linux32 muscle && rm muscle3.8.31_i86linux32.tar.gz
```
tfrcarvalho, line 103: Is this a version not available with `apt`?

consensus-genome/run.wdl
```diff
@@ -715,6 +726,30 @@ task CallVariants {
...
+    input {
+        String prefix
+        String sample
+        File ref_fasta
+        File consensus
+        String docker_image_id
+    }
+
+    command <<<
+        cat "~{consensus}" "~{ref_fasta}" > "~{sample}.muscle.in.fasta"
```
tfrcarvalho, line 739: Just curious: Do you know how does this work exactly? We just send the consensus concatenated with the references? 
Makes me think of where details/documentation for these steps are added for beginners. :)

tfrcarvalho (commented):
Overall implementation looks good, although I wanted to understand the tool a bit better.
Since you added it to the zip, it will be available to the end user, right?

consensus-genome/run.wdl
katrinakalantar: Yep, Muscle accepts a fasta file containing all the sequences that are to be aligned via multiple sequence alignment (so in this case we want to align the reference and the consensus genome) and then outputs a multiple sequence alignment file. Some documentation here: https://www.drive5.com/muscle/manual/basic_alignment.html

We don't have the steps detailed in the github repo, but we do have some documentation in the help center. Do you think it would be valuable to add more documentation within the step comments?

consensus-genome/Dockerfile
katrinakalantar: Ah, good point - it seems it is available with `apt` so I will make that change!

consensus-genome/run.wdl
tfrcarvalho: I do think it would be useful and they are easy to maintain, since we will feel more compelled to update them as we make code changes.
On the other hand, it would be bothersome to have to maintain two places.
So, not sure, what the best solution is...

tfrcarvalho (approved):

PR #139, opened 2021-06-21
Replace RunStar in host filter

rzlim08:
This might be a bit of a heavy PR. I pulled some of the functions out of idseq-dag and put them into idseq_utils in order to make them standalone functions. As this matures, we may will want to make this into a separate repository. 

The current tests and benchmarks seem to pass, I'm still making new ones to cover the different cases.


short-read-mngs/host_filter.wdl
```diff
@@ -45,23 +45,78 @@ task RunStar {
...
-    --additional-files '{"star_genome": "~{star_genome}"}' \
-    --additional-attributes '{"output_gene_file": "reads_per_gene.star.tab", "nucleotide_type": "~{nucleotide_type}", "host_genome": "~{host_genome}", "output_metrics_file": "picard_insert_metrics.txt", "output_histogram_file": "insert_size_histogram.pdf", "output_log_file": "Log.final.out"}'
+  tar -xf "~{star_genome}"
+  # Set Params
+  SAMMODE="None"
+  SAMTYPE="None"
+
+  # Currently we always use 'GeneCounts', 
+  QUANTMODE="~{if nucleotide_type == 'RNA' then 'TranscriptomeSAM GeneCounts' else 'GeneCounts'}"
+  save-description "star_out" "~{nucleotide_type}"
```
morsecodist, line 55: This is one I am not super sure about adding to the utils. Most of the descriptions are quite static and they are tightly coupled to singular steps. It would add to the code but I think it may be more clear to add here in a python block directly like:

```
python3 <<< CODE
# your code
CODE
```

This is pretty subjective though. What are your thoughts.

morsecodist (commented):
This is a really excellent change! I love the establishment of the utils and I hope to build more on this going forward. Maybe even move it so it can be used by all of our pipelines. I want to have a little discussion before merging but I am not requesting changes per se because it is mostly matters of opinion. I can't see any issues. 

One design note. I see that you are calling the utils like a binary from bash. What are your thoughts about calling them via python like?

```python
python3 <<< CODE
    from idseq_utils import foo

    foo(~{some_thing})
CODE
```

This may cut down on the boilerplate of passing things as flags and parsing. It also could make function arguments more clear. Like on this line for example:

```
save-description "star_out" "~{nucleotide_type}"
```

The caveat is that I'll admit it does look a bit more cluttered. Also I don't want us to put too much python in our bash.

Some more comments throughout.

short-read-mngs/host_filter.wdl
```diff
@@ -45,23 +45,78 @@ task RunStar {
...
+    --genomeDir STAR_genome/part-0/ \
+    --quantMode $QUANTMODE \
+    --readFilesIn "~{sep='" "' valid_input_fastq}" 
+  fi
+
+  if [ -f "Aligned.toTranscriptome.out.bam" ]; then 
+    mv "Aligned.toTranscriptome.out.bam" "Aligned.out.bam"
+  fi
+
+  sync-pairs Unmapped.out.mate*
```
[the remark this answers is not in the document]
rzlim08: Frankly I'm unsure about the necessity of this line, but I don't think I know enough about the context to say whether it is or not. Not sure if it's in scope for this ticket, but I wouldn't be opposed to replacing this line with something more meaningful.

short-read-mngs/host_filter.wdl
rzlim08: The only reason I'm hesitant about this is that some of the descriptions can be pretty long. I think we may be able to shorten them pretty significantly though. Just looking at RunSTAR, a good portion of it is describing the bash command, which is kind of redundant with miniwdl. I think it's a good idea though, and would certainly lead to more clarity since the description would be included with the code itself.

rzlim08:
> One design note. I see that you are calling the utils like a binary from bash. What are your thoughts about calling them via python like?
> 
> ```python
> python3 <<< CODE
>     from idseq_utils import foo
> 
>     foo(~{some_thing})
> CODE
> ```

I think this is a great idea! I didn't think of it, but it would probably be better than the current solution. I'll see if I can convert this.

rzlim08:
@morsecodist 

The python scripts should have been replaced by inline code here. I haven't replaced the read_descriptions util, as I think it may be better to rethink/rewrite our descriptions in general.

morsecodist (approved):
This looks great thanks!

PR #143, opened 2021-07-26
succeed without phylo tree

morsecodist:
As per our discussion, we will no longer consider it a failure if the samples are too divergent to create a phylo tree as we still have some results to show.

### DIFF
--- a/phylotree-ng/run.wdl
+++ b/phylotree-ng/run.wdl
@@ -238,6 +238,8 @@ task GenerateClusterPhylos {
 
     num_zero_variant_samples=`grep -c "^$" ska.variants.aln`
     if [[ $num_zero_variant_samples -gt 0 ]]; then
+      # If we have more than one 0 variant samples then the samples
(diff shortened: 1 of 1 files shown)


phylotree-ng/run.wdl
```diff
@@ -238,17 +238,15 @@ task GenerateClusterPhylos {
 
     num_zero_variant_samples=`grep -c "^$" ska.variants.aln`
     if [[ $num_zero_variant_samples -gt 0 ]]; then
-        export error=TooDivergentError cause="Sequences are too divergent to create a single phylo tree"
-        jq -nc ".wdl_error_message=true | .error=env.error | .cause=env.cause" > /dev/stderr
-        exit 4
+      exit 0
```
ovalenzuela19, line 241: Should we add a comment here about the reason behind the change?

morsecodist: Sure

j-x-han (approved):
👍 Thanks for doing this!

PR #145, opened 2021-07-28
Fix input min/max length when using midnight primers

rzlim08:
This changes the default inputs of our pipeline to account for inputs that use midnight primers. 

The testing of this is a bit hacky, since the change only occurs on running the entire workflow. To save time, I created a blank fastq.gz file and run that, which only takes about 30 seconds before the pipeline breaks. I am open to any suggestions on if I can do this better though.


morsecodist (approved):

katrinakalantar (commented):
This looks good to me as well!

PR #148, opened 2021-08-04
Add names to newick

morsecodist:
As discussed added the sample names to the newick file so they will be there if the user downloads.

### DIFF
--- a/phylotree-ng/run.wdl
+++ b/phylotree-ng/run.wdl
@@ -75,6 +75,15 @@ workflow phylotree {
         docker_image_id = docker_image_id
     }
 
+    if (GenerateClusterPhylos.phylotree_newick != None) {
+      call AddSampleNamesToNewick {
+        input:
+        newick = GenerateClusterPhylos.phylotree_newick,
+        samples = samples,
+        docker_image_id = docker_image_id
+      }
+    }
+
     call FetchNCBIMetadata {
         input:
         reference_accession_ids = additional_reference_accession_ids,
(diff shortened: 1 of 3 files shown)


j-x-han (approved):

phylotree-ng/run.wdl
```diff
@@ -75,6 +75,15 @@ workflow phylotree {
         docker_image_id = docker_image_id
     }
 
+    if (GenerateClusterPhylos.phylotree_newick != None) {
+      call AddSampleNamesToNewick {
+        input:
+        newick = select_first([GenerateClusterPhylos.phylotree_newick]),
```
katrinakalantar, line 81: Out of curiosity, what is the logic for using the `select_first` in this step, but not in the `AddSampleNamesToVariants` step since they both rely on outputs of the same step?

phylotree-ng/run.wdl
```diff
@@ -85,7 +94,7 @@ workflow phylotree {
         File ska_distances = AddSampleNamesToDistances.sample_name_distances
         File clustermap_png = ComputeClusters.clustermap_png
         File clustermap_svg = ComputeClusters.clustermap_svg
-        File? phylotree_newick = GenerateClusterPhylos.phylotree_newick
+        File? phylotree_newick = select_first([AddSampleNamesToNewick.phylotree_newick])
```
katrinakalantar, line 97: Related to the question above about the use of `select_first( )`, since this output previously didn't require that logic.

katrinakalantar (commented):
This looks good to me - should produce the desired output! I just had a few questions about the use of the `select_first( )` functionality.

phylotree-ng/run.wdl
morsecodist: The select first is part of the handling of chaining optional stuff. So since that output may not exist we need `select_first`. For the second one the issue is the entire step may not exist, not just the output.

PR #150, opened 2021-08-09
fix select_first for newick

morsecodist:
I think the problem is the distinction between a null input and the output of a null step. I compared this to some other examples in consensus genome and I think this is the way to do it.


jshoe (approved):
Cool, thanks! I pulled this down and it ran on this 👍

PR #152, opened 2021-08-12
handle spaces in cg sample names

morsecodist:
This will allow these file names to have spaces without breaking stuff.

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -1086,7 +1086,7 @@ task ZipOutputs {
         export TMPDIR=${TMPDIR:-/tmp}
 
         mkdir ${TMPDIR}/outputs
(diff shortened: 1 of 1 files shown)


rzlim08:
Looks good for Illumina, is it in scope to get this working for ONT pipelines? Looks like it may be more complex.

morsecodist:
@rzlim08 why so? I looked through every instance of `~{sample}` and they are all within quotes already. The only problem (I think) is that `ZipOutputs` takes arbitrary output paths from other steps and doesn't use quotes.

rzlim08:
> @rzlim08 why so? I looked through every instance of `~{sample}` and they are all within quotes already. The only problem (I think) is that `ZipOutputs` takes arbitrary output paths from other steps and doesn't use quotes.

Looks like the problem is happens when it gets passed to the artic pipeline (RunMinion), which is why it might be a bit more complex.

PR #153, opened 2021-08-12
remove another instance of TooDivergentError from phylo tree

morsecodist:
This was actually redundant with the other error. Downstream we were checking if there was more than one cluster in the directory and failing if there was. I changed that to just exiting the step but I missed this one.


j-x-han (approved):
👍  Thanks for the quick fix!

jshoe (approved):

PR #154, opened 2021-08-13
enhance csv security

morsecodist:
Add a single quote to the start of each cell that starts with a non-alphanumeric character.


rzlim08 (approved):
LGTM!

PR #157, opened 2021-08-17
force path to genome to standard name

rzlim08:
I had incorrectly assumed that the standard folder structure for the STAR genomes was "STAR_genome/part-0", when it was really "[hostname]_STAR_genome/part-0/" for everything other than the test STAR genome. 

This fix forces the directory to be "STAR_genome/part-0".


morsecodist (approved):
thanks for the quick fix!

PR #159, opened 2021-08-24
fix output gene file name

rzlim08:
reverted output gene file name


rzlim08:
Looks like the only thing failing is the notebook, which shouldn't have to do with this PR. @mlin do you know what the error is here?

jshoe (approved):
Thanks!

PR #162, opened 2021-08-30
ignore divergent iqtree error

morsecodist:
This specific error causes `iqtree` to produce no output so the `mv` command fails even though `iqtree` doesn't exit with an error code. If this error message is produced I will skip the `mv` because we want to ignore this.


j-x-han (approved):
LGTM

PR #163, opened 2021-09-02
Fix broken tests

rzlim08:
A couple of our consensus genome tests were not running what was expected, the changes to the tests were much smaller than I thought they would be.


jshoe (commented):
Hm is the notebook failure a "random" error?

rzlim08:
> Hm is the notebook failure a "random" error?
Ah, no, but the notebook failure should have been addressed in this commit: https://github.com/chanzuckerberg/idseq-workflows/commit/293655d0465645b42cfe59fdb7a2e2b72eb4d93c

jshoe (approved):
LGTM

PR #164, opened 2021-09-09
changed input parameters

rzlim08:
* reduce ivarFreqThreshold parameter from 0.9 to 0.75 and reduce minDepth parameter from 10 to 5 when using generalized consensus genome


jshoe (approved):
LGTM

PR #166, opened 2021-09-16
Relabel phylo heatmap color bar

rzlim08:
* To clarify that distances larger than 0.15 may not be displayed, I added a '>' in front of the final tick
* If we remove or change the vmin/vmax values, we will have to change the ticks as well. 
* I tried changing the vmax score, but it became hard to distinguish smaller distances


j-x-han (approved):
LGTM! Thanks for the quick fix!

PR #167, opened 2021-09-21
Fix STAR status upload

rzlim08:
Adding a comment in with the `--step-name` allows the miniwdl upload plugin to parse the step name, and upload the status to S3. We should probably rethink how we do this in the future.

### DIFF
--- a/short-read-mngs/host_filter.wdl
+++ b/short-read-mngs/host_filter.wdl
@@ -45,7 +45,8 @@ task RunStar {
     String genome_dir = "STAR_genome/part-0/"
   }
   command<<<
-  # --step-name ryan-test
+  # this comment is for the miniwdl plugin uploader to parse:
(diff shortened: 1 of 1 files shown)


jshoe (approved):
Wow!

short-read-mngs/host_filter.wdl
```diff
@@ -45,6 +45,8 @@ task RunStar {
     String genome_dir = "STAR_genome/part-0/"
   }
   command<<<
+  # this comment is for the miniwdl plugin uploader to parse:
```
tfrcarvalho, line 48: Can we create a ticket for this and add todo here with reference to the ticket?

rzlim08: Sorry I think I merged too fast. Sure thing.

PR #169, opened 2021-09-23
ignore picard failures

morsecodist:
As discussed we shouldn't fail the whole pipeline for this marginal feature. I am a bit uncomfortable with the silent failure, I added a log so we could make a metric for this. Seems to be the best option.


jshoe (approved):

rzlim08 (approved):
This is a great solution. Sorry it took so long to get to this, I wanted to confirm that the file was corrupted, and double check some other failures, which were expected given the input.

PR #170, opened 2021-09-30
Increment accession version 

rzlim08:
Solves a bug where a user requested a consensus genome with an outdated accession id.


jshoe (approved):
Sounds good to me. I wonder if we'll ever be off-by-two accessions 😆

rzlim08:
> Sounds good to me. I wonder if we'll ever be off-by-two accessions 😆

Haha I think it's possible. It looks like it's rare to be one-off, but I bet when we start updating NT more often we get into some weird edge cases.

PR #172, opened 2021-10-13
replace python script with seqkit (Go)

rzlim08:
The performance here should be greatly improved. On a test dataset (1.4GB), the step took 1m26s compared to 43m27s. 

Still need to do some testing on seqkit's error conditions. @katrinakalantar do you think we may have any issues with using seqkit? I'm mainly concerned that it may be more restrictive than previous validation steps that we have had.

### DIFF
--- a/consensus-genome/run.wdl
+++ b/consensus-genome/run.wdl
@@ -303,7 +303,7 @@ task ValidateInput{
     }
 
     command <<<
-        set -euxo pipefail 
+        set -uxo pipefail 
         function raise_error {
             export error=$1 cause=$2
             jq -nc ".wdl_error_message=true | .error=env.error | .cause=env.cause" > /dev/stderr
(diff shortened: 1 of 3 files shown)


consensus-genome/run.wdl
```diff
@@ -303,55 +303,26 @@ task ValidateInput{
...
-        import json
-        import os
-        import sys
-
-        from Bio import SeqIO
+        if [[ "~{technology}" == "ONT" ]] && [[ "~{length(fastqs)}" -gt 1 ]]; then
+            export error=InvalidInputFileError cause="Multiple fastqs provided for ONT"
+            jq -nc ".wdl_error_message=true | .error=env.error | .cause=env.cause" > /dev/stderr
+            exit 1 
+        fi
```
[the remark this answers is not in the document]
rzlim08: Yes!

consensus-genome/run.wdl
```diff
@@ -303,67 +303,54 @@ task ValidateInput{
...
-            # ONT guppyplex requires files are in .fastq format with .fastq extension (not .fq)
-            prefix = "~{prefix}"
-            out_path = f"{prefix}validated_{i}.fastq.gz" if len(fastqs) > 1 else f"{prefix}validated.fastq.gz"
-            with maybe_gzip_open(file) as in_f, gzip.open(out_path, "wt") as out_f:
-                SeqIO.write(truncate(SeqIO.parse(in_f, "fastq")), out_f, "fastq")
-        CODE
+        counter=1
+        for fastq in ~{sep=' ' fastqs}; do 
+           # limit max # of reads to max_reads
+           seqkit head -n "~{max_reads}" $fastq -o "~{prefix}validated_$counter.fastq.gz"
```
morsecodist, line 339: Shouldn't this go first? You are going to be computing stats on the whole file just to take the head. I guess you could argue we would want to reject the file if it is invalid anywhere but I am not sure that is necessary.

rzlim08: Hm, you're probably right, I was thinking it might be good to count the total number of reads just to see if we do end up subsampling, how much we do so by, but it also doesn't really make sense to fail the whole sample if the 75,000,001th read has 301 bps or whatever. I'll make the change!

morsecodist (approved):
Thanks!

PR #173, opened 2021-10-15
Separate step and integration tests

rzlim08:
Separated the step and integration tests for consensus genomes, added the integration tests to self-hosted runners. 

Time to run tests is down from ~22 minutes to ~8 minutes. It probably won't get much lower as a lot of the time now is for setup. 

Also taking advantage of caching for the integration tests.


morsecodist (approved):
Looks great!

PR #180, opened 2021-11-01
fix STAR known errors by printing error JSON

morsecodist:
In order for the runner script to detect errors they must be printed as JSON as the last line of stderr. Maybe this constraint is not ideal but I think the best way to solve it for now is by printing errors. I took the code from our running of single pipeline steps:

```python
    try:
        if args.step_class == "PipelineStepRunValidateInput":
            count_input_reads(input_files=step_instance.input_files_local,
                              max_fragments=step_instance.additional_attributes["truncate_fragments_to"])

        step_instance.validate_input_files()
        with open(f"{args.step_name}.description.md", "wb") as outfile:
            # write step_description (which subclasses may generate dynamically) to local file
            outfile.write(step_instance.step_description().encode("utf-8"))
        step_instance.run()
        step_instance.count_reads()
        step_instance.save_counts()
    except Exception as e:
        traceback.print_exc()
        exit(json.dumps(dict(wdl_error_message=True, error=type(e).__name__, cause=str(e), step_description_md=step_instance.step_description())))
```

and adapted it for our new utils.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ In addition, if a modification is made that may affect the results of a pipeline
 
 ### Unreleased
 
+### short-read-mngs-v6.8.7
+- fix bug in RunSTAR that caused known errors not to be reported
+
 ### consensus-genome-v3.4.4
 - raise limit for illumina CGs to 500 basepairs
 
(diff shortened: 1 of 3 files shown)


rzlim08:
This looks good, thanks for this! Yeah it seems like a challenge to convert the python exceptions to JSON, but this looks like a pretty good solution.

rzlim08 (approved):

PR #184, opened 2021-11-12
Add minimap2 to wdl 

rzlim08:
This builds on top of the previous pull request, and adds minimap2 to the non-host alignment wdl but hides it behind a flag.


rzlim08:
@morsecodist I think this is ready to look at.

